PR #1, opened 2022-10-10
Fixes #RHIROS-851 - Listener for consuming cost-mgmt events.

patilsuraj767:


ros/lib/logger.py
```diff
@@ -0,0 +1,80 @@
...
+import socket
+from threading import local
+from ros.lib.config import LOG_LEVEL
+from logstash_formatter import LogstashFormatterV1
+
+
+threadctx = local()
+
+
+def cloudwatch_config_values():
```
kgaikwad, line 13: If I understand correctly, this is initial version and it includes local setup details & not deployment change, does it make sense to include cloudwatch configs and clowder handling in this PR?

kgaikwad (changes requested):
@patilsuraj767 - added a single inline comment otherwise we can go ahead with rest changes.

Makefile
```diff
@@ -0,0 +1,10 @@
+identity={"identity": {"org_id": "3340851", "type": "System", "auth_type": "cert-auth", "system": {"cn": "1b36b20f-7fa0-4454-a6d2-008294e06378", "cert_type": "system"}, "internal": {"org_id": "3340851", "auth_time": 6300}}}
+b64_identity=$(shell echo '${identity}' | base64 -w 0 -)
+
+file=./scripts/samples/cost-mgmt.tar.gz
+
+local-upload-data:
+	curl -vvvv -F "upload=@$(file);type=application/application/vnd.redhat.hccm.tar+tgz" \
+		-H "x-rh-identity: ${b64_identity}" \
+		-H "x-rh-request_id: testtesttest" \
+		http://localhost:3000/api/ingress/v1/upload
```
upadhyeammit, line 10: Can we also get the port from the ENV var?

patilsuraj767: Yes, we can but is there any use case for it? We have hard coded ingress port to 3000 in docker-compose, if we add env variable then we will also need to change docker-compose.yaml.

upadhyeammit: The use case is same, I have another services running on the same port and instead of editing code to run the setup I can export env var?

upadhyeammit (commented):
Overall should we change directory structure to `ros-ocp` or `ros_ocp`?

Makefile
patilsuraj767: Done, I have made ingress port configurable.

ros/lib/config.py
```diff
@@ -0,0 +1,40 @@
...
+
+LOG = logging.getLogger(__name__)
+
+CLOWDER_ENABLED = True if os.getenv("CLOWDER_ENABLED", default="False").lower() in ["true", "t", "yes", "y"] else False
+
+
+def kafka_auth_config(connection_object):
+    if KAFKA_BROKER:
+        if KAFKA_BROKER.cacert:
+            connection_object["ssl.ca.location"] = "/tmp/cacert"
```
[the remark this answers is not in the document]
patilsuraj767: Regrading module name convention "-" is not allowed and use of "_" (underscores) is discouraged [1]. Will you be OK if I name module rosocp? 
Regarding filename of kafka cacert I don't think it will be required anywhere apart from creating kafka connection object and I think we can make it as constant in future if any such need arise. Let me know your thoughts. 

[1] https://peps.python.org/pep-0008/#package-and-module-names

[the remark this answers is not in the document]
upadhyeammit: Thanks for the reference on naming convention, I always forget to check that :) . If I understand correctly by below para then underscores are only discouraged in package names? so in that case I would choose ros_ocp,
~~~
Modules should have short, all-lowercase names. Underscores can be used in the module name if it improves readability.
Python packages should also have short, all-lowercase names, although the use of underscores is discouraged.
~~~

For the ca cert file, I can see we referred that twice and so assumed we can have constant for that, sometimes my mind tries to avoid as much errors as we can using the DRY principles :)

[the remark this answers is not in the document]
patilsuraj767: Done, I have changed the package name and declared kafka ca cert path as constant.

kgaikwad (approved):
LGTM  as first version

upadhyeammit (approved):

patilsuraj767:
Thanks @upadhyeammit and @kgaikwad for the review. Merging the PR.

PR #12, opened 2023-02-28
Fixes #RHIROS-966 - CSV data aggregation

patilsuraj767:


cmd/aggregator.go
```diff
@@ -0,0 +1,66 @@
...
+				if _, err := os.Stat(outputDir); os.IsNotExist(err) {
+					if err := os.MkdirAll("a/b/c/d", os.ModePerm); err != nil {
+						panic(err.Error())
+					}
+				}
+			} else {
+				outputDir, _ = os.Getwd()
+			}
+			f, err := os.Open(input_file)
+			if err != nil {
```
upadhyeammit, line 36: Can we extract this if block which is repetitive as function and call?

cmd/aggregator.go
```diff
@@ -0,0 +1,66 @@
...
+				panic(err.Error())
+			}
+			defer f.Close()
```
upadhyeammit, line 39: Does this mean we keep the f open when there is error?

cmd/aggregator.go
```diff
@@ -0,0 +1,66 @@
...
+
+import (
+	"encoding/csv"
+	"fmt"
+	"os"
+
+	"github.com/go-gota/gota/dataframe"
+	"github.com/spf13/cobra"
+
+	"github.com/redhatinsights/ros-ocp-backend/internal/processor"
```
upadhyeammit, line 11: Can this be converted to local import?
something like `internal/processor`?

cmd/aggregator.go
```diff
@@ -0,0 +1,66 @@
...
+
+			csv := csv.NewReader(f)
+			records, err := csv.ReadAll()
+			if err != nil {
+				panic(err.Error())
+			}
+
+			df := dataframe.LoadRecords(records)
+			df = processor.Aggregate_data(df)
+			fileio, err := os.Create(outputDir + "/output.csv")
```
upadhyeammit, line 49: Do you think we should have var for output.csv?

upadhyeammit (commented):
With my limited knowledge added few comments, request to check.

cmd/aggregator.go
patilsuraj767: No, `f.Close` first checks if `f` is nil - https://cs.opensource.google/go/go/+/refs/tags/go1.20.1:src/os/file_posix.go;l=21 and then perform the action.

cmd/aggregator.go
patilsuraj767: No, we can't. This is how exception handling done in idiomatic Go style.

cmd/aggregator.go
patilsuraj767: ok I will create var `outputFile := outputDir + "/output.csv"`

internal/processor/aggregator.go
```diff
@@ -0,0 +1,85 @@
...
+
+	s := df.Rapply(func(s series.Series) series.Series {
+		owner_name := s.Elem(index_of_owner_name).String()
+		owner_kind := s.Elem(index_of_owner_kind).String()
+		workload := s.Elem(index_of_workload).String()
+		workload_type := s.Elem(index_of_workload_type).String()
+		if owner_kind == "ReplicaSet" && workload == "<none>" {
+			return series.Strings([]string{"replicaset", owner_name})
+		} else if owner_kind == "ReplicationController" && workload == "<none>" {
+			return series.Strings([]string{"replicationcontroller", owner_name})
```
saltgen, line 31: Shall we use the WorkloadType enum values here?

internal/processor/kruize_api.go
```diff
@@ -2,169 +2,108 @@ package processor
...
-		return errors.New(error_string)
+		return fmt.Errorf("an Error Occured %v", err)
 	}
 	q := req.URL.Query()
-	q.Add("experiment_name", params["experiment_name"])
-	q.Add("deployment_name", params["deployment_name"])
-	q.Add("namespace", params["namespace"])
+	q.Add("experiment_name", experiment)
+	req.URL.RawQuery = q.Encode()
 	res, err := client.Do(req)
```
saltgen, line 104: Is the intention here to have recommendations saved to DB post the Recommender system is introduced?

saltgen (approved):
@patilsuraj767 LGTM :+1: 
Just a suggestion and a question for my understanding.

internal/processor/aggregator.go
patilsuraj767: Done

internal/processor/kruize_api.go
patilsuraj767: Yes, Once recommender is implemented we will save the data coming from listRecommendation API to ROS-OCP DB.

PR #24, opened 2023-03-23
Fixes #RHIROS-1024 - Added RBAC and identity middleware

patilsuraj767:


patilsuraj767:
Below the sample output from rbac. 
```
{
  "meta": {
    "count": 1,
    "limit": 10,
    "offset": 0
  },
  "links": {
    "first": "/api/rbac/v1/access/?application=cost-management&limit=10&offset=0&username=notifications-user",
    "next": null,
    "previous": null,
    "last": "/api/rbac/v1/access/?application=cost-management&limit=10&offset=0&username=notifications-user"
  },
  "data": [
    {
      "resourceDefinitions": [

        {
          "attributeFilter": {
            "key": "cost-management.openshift.project",
            "value": [
              "openshift-qe_test"
            ],
            "operation": "in"
          }
        }

      ],
      "permission": "cost-management:openshift.project:read"
    }
  ]
}
```

internal/api/handlers.go
```diff
@@ -39,6 +35,13 @@ func GetRecommendationSetList(c echo.Context) error {
 		}
 	}
 
+	if !is_user_authorized_for_resource(types.ResourceObject{
+		Cluster: c.QueryParam("cluster"),
+		Project: c.QueryParam("project"),
+	}, user_permissions) {
```
kgaikwad, line 41: How does this condition work when user is having cluster permissions but not having permissions for few projects?  
Does that mean user can get recommendations for projects when he is having permissions for cluster, correct?

internal/api/handlers.go
```diff
@@ -50,17 +53,22 @@ func GetRecommendationSetList(c echo.Context) error {
...
+			recommendationData := make(map[string]interface{})
+			recommendationData["id"] = recommendation.ID
+			recommendationData["cluster_uuid"] = recommendation.Workload.Cluster.ClusterUUID
+			recommendationData["cluster_alias"] = recommendation.Workload.Cluster.ClusterAlias
+			recommendationData["project"] = recommendation.Workload.Namespace
+			recommendationData["workload_type"] = recommendation.Workload.WorkloadType
+			recommendationData["workload"] = recommendation.Workload.WorkloadName
+			recommendationData["container"] = recommendation.ContainerName
+			recommendationData["last_report"] = recommendation.Workload.Cluster.LastReportedAtStr
+			recommendationData["values"] = recommendation.Recommendations
```
kgaikwad, line 69: Not now but will it be possible to have structure of recommendationData so that it would be useful to initialize structure objects?

internal/api/middleware/identity.go
```diff
@@ -0,0 +1,26 @@
...
+)
+
+func Identity(next echo.HandlerFunc) echo.HandlerFunc {
+	return func(c echo.Context) error {
+		id := identity.XRHID{}
+		encodedIdentity := c.Request().Header.Get("X-Rh-Identity")
+		decodedIdentity, err := base64.StdEncoding.DecodeString(encodedIdentity)
+		if err != nil {
+			return echo.NewHTTPError(http.StatusUnauthorized, "Unable to decode X-Rh-Identity")
+		}
```
kgaikwad, line 19: After importing module `github.com/redhatinsights/platform-go-middlewares/identity`,  why did you require this method?

internal/api/middleware/rbac.go
```diff
@@ -0,0 +1,96 @@
...
+	"github.com/redhatinsights/ros-ocp-backend/internal/types"
+	"github.com/sirupsen/logrus"
+)
+
+var cfg *config.Config = config.GetConfig()
+var log *logrus.Logger = logging.GetLogger()
+
+func Rbac(next echo.HandlerFunc) echo.HandlerFunc {
+	return func(c echo.Context) error {
+		permissions := get_user_permissions(c.Request().Header.Get("X-Rh-Identity"))
```
kgaikwad, line 23: Two methods with same naming i.e `get_user_permissions` and different set of arguments (one accepts context whereas other accepts identity header) in two files. Could you please rename those to avoid confusion or make it single method to address both requirements?
Thoughts?

internal/api/middleware/rbac.go
```diff
@@ -0,0 +1,96 @@
...
+		if permissions != nil {
+			c.Set("user.permissions", permissions)
+		} else {
+			return echo.NewHTTPError(http.StatusUnauthorized, "User is not authorized")
+		}
+		return next(c)
+	}
+}
+
+func aggregate_permissions(acls []types.RbacData) map[string][]string {
```
kgaikwad, line 33: Will it be possible to add doc string with short description what this method does?

internal/config/config.go
```diff
@@ -77,6 +83,16 @@ func initConfig() {
 		viper.SetDefault("DBPort", c.Database.Port)
 		viper.SetDefault("DBssl", c.Database.SslMode)
 
+		// clowder RBAC Config
+		for _, endpoint := range c.Endpoints {
+			if endpoint.App == "rbac" {
+				viper.SetDefault("RBACHost", endpoint.Hostname)
+				viper.SetDefault("RBACPort", endpoint.Port)
+				viper.SetDefault("RBACProtocol", "https")
```
kgaikwad, line 91: For any external communication, please go through this ticket - https://issues.redhat.com/browse/RHINENG-103. Please make sure you are using correct configurations from clowder.

kgaikwad (changes requested):
@patilsuraj767  - Thank you for introducing RBAC permissions check. Added few inline comments.

internal/api/handlers.go
patilsuraj767: Yes, that's correct. If user have the cluster permission that means user is authorized for all the projects in that cluster.

internal/api/middleware/identity.go
patilsuraj767: That middleware is for golang standard libaray net/http and not for the echo framework. I have imported that package just for getting the [identity struct](https://github.com/RedHatInsights/platform-go-middlewares/blob/master/identity/identity.go#L57).

internal/config/config.go
patilsuraj767: Yes we have taken care of that. For golang application we just need to add [SSL_CERT_DIR](https://github.com/RedHatInsights/ros-ocp-backend/blob/main/clowdapp.yaml#L210) environment variable.

internal/api/middleware/rbac.go
patilsuraj767: Done. Changed the function name to `get_user_permissions_from_rbac`.

internal/api/middleware/rbac.go
patilsuraj767: Added the doc string.

internal/api/handlers.go
saltgen: @patilsuraj767 In that case, do we need L40?

internal/api/server.go
```diff
@@ -5,23 +5,33 @@ import (
...
+	app.GET("/status", GetAppStatus)
+
+	v1 := app.Group("/api/cost-management/v1")
+	v1.Use(ros_middleware.Identity)
+	if cfg.RBACEnabled {
+		v1.Use(ros_middleware.Rbac)
+	}
+	v1.GET("/recommendations/openshift", GetRecommendationSetList)
+	v1.GET("/recommendations/openshift/:recommendation-id", GetRecommendationSet)
+	v1.File("/recommendations/openshift/openapi.json", "openapi.json")
```
saltgen, line 34:
```suggestion
	v1.File("/recommendations/openshift/openapi.json", "internal/api/docs/v1/openapi.json")
```

Getting following error above change, I think we should exclude this endpoint from rbac

`{
    "message": "Unable to marshal X-Rh-Identity into struct"
}`

kgaikwad: @saltgen  +1 to excluding this endpoint.  URL `{path} /openapi.json` should be externally available.

internal/api/middleware/rbac.go
```diff
@@ -0,0 +1,102 @@
...
+			permissions["*"] = []string{}
+		}
+	}
+	return permissions
+}
+
+func get_user_permissions_from_rbac(encodedIdentity string) map[string][]string {
+	cfg := config.GetConfig()
+	url := fmt.Sprintf(
+		"%s://%s:%s/api/rbac/v1/access/?application=cost-management&limit=100",
```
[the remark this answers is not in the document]
patilsuraj767: `count`? I didn't get you which `count` you are talking about?

internal/api/server.go
patilsuraj767: `/openapi.json` does not require any RBAC permission but atleast user should be logged into consoledot.

internal/api/middleware/rbac.go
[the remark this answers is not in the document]
saltgen: The count for the rbac url, from the sample above.
https://github.com/RedHatInsights/ros-ocp-backend/pull/24#issuecomment-1482708944

internal/api/server.go
saltgen: IIUC, console login should not be a necessity as well. I get valid response from below request.
`curl "https://console.redhat.com/api/policies/v1/openapi.json" | python -m json.tool `

patilsuraj767: Ok I will exclude /openapi.json route from rbac.

internal/api/middleware/rbac.go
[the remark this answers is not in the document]
patilsuraj767: `count` refer to number of permissions under data section which are returned with rbac API response. I am not sure what is the default set by rbac.

internal/api/handlers.go
patilsuraj767: Yes, we need it. Because it may happen user do not have any cluster permission but he has project level permission.

internal/api/utils.go
```diff
@@ -141,3 +142,43 @@ func MapQueryParameters(c echo.Context) map[string]interface{} {
...
+		if resourceObject.Cluster == "" && resourceObject.Project == "" {
+			return true
+		}
+		if stringInSlice(resourceObject.Cluster, user_permissions["openshift.cluster"]) ||
+			stringInSlice(resourceObject.Project, user_permissions["openshift.project"]) ||
+			stringInSlice("*", user_permissions["openshift.cluster"]) ||
+			stringInSlice("*", user_permissions["openshift.project"]) {
+			return true
+		}
+	} else {
```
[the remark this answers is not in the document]
patilsuraj767: else block cover the scenario when RBAC is disabled. Mostly used in local setup when we disable RBAC using env variable.

saltgen (approved):
@patilsuraj767 Thanks for the changes LGTM :+1:

PR #25, opened 2023-03-28
Refs RHIROS-1014 - DB changes required for stage deployment

patilsuraj767:


patilsuraj767:
This PR is now invalid, because now we are going further with the separate rds for ROS-OCP hence we don't require separate schema and search_path.

PR #40, opened 2023-04-04
Check if performance profile is created before uploading metric

patilsuraj767:


saltgen (approved):
Thanks @patilsuraj767!

PR #41, opened 2023-04-04
Fix DB constraint

patilsuraj767:


patilsuraj767:
Just reverted the API changes and DB constraint which was added in earlier PR.

PR #42, opened 2023-04-04
Updated sample archive data

patilsuraj767:


pgarciaq:
## Implementation Plan: Option A — `is_stale` column on `org_container_keys`

### Chosen Approach

**Option A** from the issue analysis: Add an `is_stale BOOLEAN NOT NULL DEFAULT false` column to `org_container_keys`, so ALL containers (stale and non-stale) live in the keys table. This lets every stale filter variant (`false`, `true`, `only`) use the same fast keyset-pagination path through `org_container_keys`.

### Correctness Bug Discovered

During analysis, a critical correctness bug was identified in `getNativeRecommendationsDistinct`:

```go
distinctSubquery := db.Table("recommendation_sets rs").
    // ...
    Where("rs.stale = false")  // ← hardcoded regardless of filter[stale]
```

This means:
- `filter[stale]=true` (show both stale and non-stale) → the hardcoded `WHERE rs.stale = false` silently discards stale rows, producing the **same results as the default**
- `filter[stale]=only` (show only stale) → `queryParams` adds `WHERE rs.stale = true` **on top of** the hardcoded `WHERE rs.stale = false`, producing **zero rows**

Both are incorrect. The fix is part of this implementation — once all queries go through the `org_container_keys` path, the broken `getNativeRecommendationsDistinct` stale handling becomes moot.

### Implementation Steps

1. **Migration 000154**: Add `is_stale BOOLEAN NOT NULL DEFAULT false` to `org_container_keys`, with a partial index `ON org_container_keys (org_id) WHERE is_stale = true`
2. **`RefreshOrgContainerKeysTx`**: Upsert ALL containers (not just `stale = false`), setting `is_stale = rs.stale`. Remove the DELETE step (stale containers stay in the table with `is_stale = true`)
3. **`usesOrgContainerKeys()`**: Always return `true` when the stale key is present (regardless of value), so all three filter modes use the keys path
4. **`buildNativeContainerKeysPageQuery`**: Apply `WHERE ock.is_stale = ?` based on the filter, and make the RS join stale-aware
5. **`nativeContainerRSJoin`**: Parameterize (or remove) the hardcoded `AND rs.stale = false` for stale-inclusive queries
6. **Count queries**: Make count queries stale-aware via the same `ock.is_stale` predicate
7. **Tests**: New tests for all three stale filter paths via org_container_keys, plus refresh correctness with `is_stale`
8. **ADR, OpenAPI spec, CHANGELOG updates**

### Trade-offs

- **+** Unified query path = simpler code, fewer bugs, consistent pagination
- **+** Fixes broken `filter[stale]=true` and `filter[stale]=only`
- **−** `org_container_keys` table grows (includes stale containers), but stale containers are typically <10% of total
- **−** Partial index needed for efficient `filter[stale]=only` queries

PR #45, opened 2023-04-06
RHIROS-1058 Fix count in API List response

saltgen:
The count value currently returns the number of records present in API response.

This should be the actual number of records on database with filters (if present).


patilsuraj767 (approved):
LGTM :+1:

PR #46, opened 2023-04-11
Fixes #RHIROS-1062 - Send 6hrs of data in chronological order

patilsuraj767:


internal/processor/utils.go
```diff
@@ -108,6 +108,15 @@ func convertDateToISO8601(date string) string {
 	return t.Format("2006-01-02T15:04:05.000Z")
 }
 
+func convertStringToTime(data string) (time.Time, error) {
```
r14chandra, line 111: Can we add a test for `convertStringToTime` util function?

r14chandra (changes requested):

r14chandra (approved):
Looks good!

PR #50, opened 2023-04-17
Fixes #RHIROS-1078 - Added new column for source_id

patilsuraj767:


Makefile
```diff
@@ -100,5 +100,5 @@ ifdef env
 else
 	curl -H "x-rh-identity: ${b64_identity}" \
 		 -H "x-rh-request_id: testtesttest" \
-		 http://localhost:8000/api/cost-management/v1/recommendations/openshift
-endif
\ No newline at end of file
+		 http://localhost:8000/api/cost-management/v1/recommendations/openshift | jq
```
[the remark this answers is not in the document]
patilsuraj767: I don't think it is by default installed in users system.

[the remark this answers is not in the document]
r14chandra: Then can we use `json.tool` here? `python -m json.tool`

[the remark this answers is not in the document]
patilsuraj767: Yes, we can do that. I will make the changes.

internal/api/handlers.go
```diff
@@ -190,6 +191,7 @@ func GetRecommendationSet(c echo.Context) error {
 		longTermSection["variation"] = variationDummyObject
 
 		recommendationSlice["id"] = recommendationSet.ID
+		recommendationSlice["source_id"] = recommendationSet.Workload.Cluster.SourceId
```
saltgen, line 194: Do we need source_id in API responses?

saltgen (changes requested):

r14chandra (approved):
LGTM!

internal/api/handlers.go
patilsuraj767: Not sure, I thought QE might need for verification so added it.

saltgen: Posted a question in the Arch doc for Chris regarding the same

saltgen: This is good to go, change approved on doc

saltgen (approved):

PR #53, opened 2023-04-19
RHIROS-1081  Misc openapi updates

saltgen:
- Redefined url paths in openapi file
- Introduced duration_based, and all duration samples(short-term, long-term, medium-term) in schema
- Updated value type for Recommendation id
- Minor text updates
- Updated links section with `previous` and `next`
- Removed `minimum` attribute for recommendation-id path parameter


dgaikwad:
Now I am able to get responses from APIs, LGTM. ACK to merge by me.

internal/api/docs/v1/openapi.json
```diff
@@ -216,120 +215,511 @@
...
+          "cluster_uuid": {
             "type": "string",
-            "example": "proj_rxu"
+            "example": "d29c4b8b-f1a8-471c-ab95-b64e36bb51a9"
           },
-          "workload": {
+          "container": {
             "type": "string",
-            "example": "deployment_proj_rxu"
+            "example": "pod-ros-A11"
```
kgaikwad, line 228:
```suggestion
            "example": "pod-ros-A11"
```
do we really need to provide examples here for fields like cluster_uuid, container, cluster_alias for Recommendation object?

kgaikwad (approved):
@saltgen,  just added one inline comment for my understanding otherwise looks good. :+1:

internal/api/docs/v1/openapi.json
saltgen: @kgaikwad Yes, if we add the example values along with the component schema it generates the sample response automatically

PR #69, opened 2023-04-28
RHIROS-1102 Kruize - Experiment creation failure

saltgen:


internal/utils/kruize/kruize_api.go
```diff
@@ -45,14 +45,36 @@ func Create_kruize_experiments(experiment_name string, k8s_object []map[string]i
...
+		log.Error("Performance profile does not exist")
+		log.Info("Tring to create resource_optimization_openshift performance profile")
+		utils.Setup_kruize_performance_profile()
+	}
+
+	// Re-attempting experiment creation request
+	res, err = http.Post(url, "application/json", bytes.NewBuffer(payload))
+	if err != nil {
+		return nil, fmt.Errorf("error Occured while creating experiment: %v", err)
+	}
```
patilsuraj767, line 69: This will duplicate every create experiment request.

patilsuraj767 (changes requested):

internal/utils/kruize/kruize_api.go
patilsuraj767: Can we check the response of create experiment and if it says `Performance Profile doesn't exist` then create the performance profile and give recursive call to `Create_kruize_experiments` function.

internal/utils/kruize/kruize_api.go
```diff
@@ -53,6 +53,16 @@ func Create_kruize_experiments(experiment_name string, k8s_object []map[string]i
...
 		return nil, fmt.Errorf("can not unmarshal response data: %v", err)
 	}
+
+	// Temporary fix
+	// Currently, once Kruize pod inits it does not load performance-profile from DB
+	if strings.Contains(resdata["message"].(string), "Performance Profile doesn't exist") {
+		log.Error("Performance profile does not exist")
+		log.Info("Tring to create resource_optimization_openshift performance profile")
+		utils.Setup_kruize_performance_profile()
+		_, _ = Create_kruize_experiments(experiment_name, k8s_object)
```
patilsuraj767, line 63: We need to return the response to main calling function. 

```suggestion
		if container_names, err := Create_kruize_experiments(experiment_name, k8s_object); err != nil {
				return nil, err
			} else {
				return container_names, nil
			}
```

patilsuraj767 (changes requested):

internal/utils/kruize/kruize_api.go
```diff
@@ -45,14 +46,25 @@ func Create_kruize_experiments(experiment_name string, k8s_object []map[string]i
 	}
 	res, err := http.Post(url, "application/json", bytes.NewBuffer(payload))
 	if err != nil {
-		return nil, fmt.Errorf("error Occured while creating experiment: %v", err)
+		log.Errorf("error Occured while creating experiment: %v", err)
```
patilsuraj767, line 49: We need to `return` from here if there is any issue while calling kruize API.

saltgen: Won't this exit the function?

saltgen: Ah, for other scenarios, you mean

patilsuraj767: When kruize return "performance profile already exist" at that time this is not triggered. err is only set when there is any issue in making http request. for example - kruize is down.

PR #73, opened 2023-05-09
Fixes #RHIROS-1121 - omit metrics object if values not present

patilsuraj767:
Kruize expects that if user has not set the cpu/memory requests and limits then we should omit that object from json payload. 
Currently we are sending -1.

### DIFF
--- a/cmd/validate.go
+++ b/cmd/validate.go
@@ -1,0 +1,69 @@
+package cmd
+
+import (
+	"encoding/csv"
+	"fmt"
+	"os"
+
+	"github.com/go-gota/gota/dataframe"
+	"github.com/spf13/cobra"
+)
+
+var validateCmd = &cobra.Command{Use: "validate", Short: "Validate ros related data"}
+
+var validateCSV = &cobra.Command{
+	Use:   "csv [input csv file path]",
+	Short: "Validate ros-usage CSV file",
+	Long:  "Validate ros-usage CSV file",
+	Run: func(cmd *cobra.Command, args []string) {
+		input_file := args[0]
+		if _, err := os.Stat(input_file); os.IsNotExist(err) {
+			fmt.Printf("CSV file: %s does not exist\n", input_file)
+			os.Exit(1)
+		}
+		f, err := os.Open(input_file)
+		if err != nil {
+			panic(err.Error())
+		}
+		defer f.Close()
+
+		csv := csv.NewReader(f)
+		records, err := csv.ReadAll()
+		if err != nil {
+			panic(err.Error())
+		}
+
+		df := dataframe.LoadRecords(records)
+		groups := df.GroupBy("container_name", "pod", "owner_name", "owner_kind", "workload", "workload_type", "namespace", "image_name").GetGroups()
+
+		// for _, v := range groups {
+		// 	if v.Nrow() > 26 {
+		// 		sorted := v.Arrange(
+		// 			dataframe.Sort("interval_start"),
+		// 		)
+		// 		for _, data := range sorted.Maps() {
+		// 			asd, _ := utils.ConvertStringToTime(data["interval_start"].(string))
+		// 			fmt.Println(asd)
+		// 		}
+		// 	}
+		// }
+
+		valid_experiments := 0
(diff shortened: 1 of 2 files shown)


internal/types/kruizePayload/common.go
```diff
@@ -91,103 +91,160 @@ func convertMetricToString(data interface{}) string {
...
-						Avg:    convertMetricToString(c["memory_rss_usage_container_avg_MEAN"]),
-						Format: "MiB",
-					},
+		})
+	}
+
+	// memoryRSS
+	sum = convertMetricToString(c["memory_rss_usage_container_sum_SUM"])
+	avg = convertMetricToString(c["memory_rss_usage_container_avg_MEAN"])
+	max = convertMetricToString(c["memory_rss_usage_container_min_MIN"])
```
bharathappali, line 227: Please correct me if I'm wrong. I would expect `max` to be mapped with something like `memory_rss_usage_container_max_MAX` as I found out that `memory_rss_usage_container_min_MIN` is already being used for `min`, Also in the earlier code the `max` was set to the return value of this `convertMetricToString(c["memory_rss_usage_container_max_MAX"])`

internal/types/kruizePayload/common.go
```diff
@@ -91,103 +91,160 @@ func convertMetricToString(data interface{}) string {
...
-						Format: "cores",
-					},
+		})
+	}
+
+	// cpuUsage
+	sum = convertMetricToString(c["cpu_usage_container_sum_SUM"])
+	avg = convertMetricToString(c["cpu_usage_container_avg_MEAN"])
+	max := convertMetricToString(c["cpu_usage_container_max_MAX"])
+	min := convertMetricToString(c["cpu_usage_container_min_MIN"])
```
saltgen, line 138: @patilsuraj767 Shall we check for max and min as well on the if condition for all metric appends?

internal/types/kruizePayload/common.go
patilsuraj767: Oh, that's correct. Thanks for pointing it out. I will make the changes.

internal/types/kruizePayload/common.go
```diff
@@ -91,103 +91,160 @@ func convertMetricToString(data interface{}) string {
...
-						Format: "MiB",
-					},
+		})
+	}
+
+	// memoryUsage
+	sum = convertMetricToString(c["memory_usage_container_sum_SUM"])
+	avg = convertMetricToString(c["memory_usage_container_avg_MEAN"])
+	min = convertMetricToString(c["memory_usage_container_min_MIN"])
+	max = convertMetricToString(c["memory_usage_container_max_MAX"])
```
saltgen, line 208: @patilsuraj767 Shall we check for `min` and `max` on if conditions for each metric append?

patilsuraj767: I just thought that it might happen for some reason we might have `sum` and `avg` but do not have `min` and `max`.  As far as I know sum and avg is a mandatory so I added condition only on `sum` and `avg`.

saltgen (approved):
LGTM :+1:

cmd/validate.go
```diff
@@ -0,0 +1,69 @@
...
+		// 	if v.Nrow() > 26 {
+		// 		sorted := v.Arrange(
+		// 			dataframe.Sort("interval_start"),
+		// 		)
+		// 		for _, data := range sorted.Maps() {
+		// 			asd, _ := utils.ConvertStringToTime(data["interval_start"].(string))
+		// 			fmt.Println(asd)
+		// 		}
+		// 	}
+		// }
```
kgaikwad, line 49: @patilsuraj767  - Though the pull-request is merged, are these unwanted leftovers?
In case intentional, as nitpick could you add parent comment with keyword (example - //TODO or //FIXME) that explains reason behind keeping this code. Otherwise, does it make sense to remove these lines?

patilsuraj767: @kgaikwad that was not intentional. I was something else on which I am working on. I have created the PR to remove that file - https://github.com/RedHatInsights/ros-ocp-backend/pull/75

PR #77, opened 2023-05-10
Fixes #RHIROS-1123 - check if container recommendation is valid

patilsuraj767:
Patch for - https://github.com/kruize/autotune/issues/770

### DIFF
--- a/internal/services/recommender.go
+++ b/internal/services/recommender.go
@@ -57,10 +57,12 @@ func ProcessEvent(msg *kafka.Message) {
 
 				notifications := v.Duration_based.Short_term.Notifications
 
-				// Below "if" condition is the kruize patch for - https://github.com/kruize/autotune/issues/770
-				// issue - RHIROS-1123
-				// Should be removed once fixed in kruize
-				if !(len(notifications) > 1 && notifications[0].Notificationtype == "info" && notifications[0].Message == "There is not enough data available to generate a recommendation.") {
+				// Note - added patch here to handle kruize side issue
+				// Kruize issue: https://github.com/kruize/autotune/issues/770
+				// ROS issue - RHIROS-1123
+				// TODO: Should be removed once fixed in kruize
+				KRUIZE_NO_SUFFICIENT_DATA := "There is not enough data available to generate a recommendation."
+				if !(len(notifications) > 0 && notifications[0].Notificationtype == "info" && notifications[0].Message == KRUIZE_NO_SUFFICIENT_DATA) {
 					marshalData, err := json.Marshal(v)
 					if err != nil {
 						log.Errorf("Unable to list recommendation for: %v", err)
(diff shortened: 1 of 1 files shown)


internal/services/recommender.go
```diff
@@ -54,7 +54,13 @@ func ProcessEvent(msg *kafka.Message) {
 		containers := data[0].Kubernetes_objects[0].Containers
 		for _, container := range containers {
 			for _, v := range container.Recommendations.Data {
-				if len(v.Duration_based.Short_term.Notifications) == 0 {
+
+				notifications := v.Duration_based.Short_term.Notifications
+
+				// Below "if" condition is the kruize patch for - https://github.com/kruize/autotune/issues/770
+				// issue - RHIROS-1123
+				// Should be removed once fixed in kruize
```
kgaikwad, line 62:
```suggestion
				// Note - added patch here to handle kruize side issue
                                // Kruize issue: https://github.com/kruize/autotune/issues/770
                                // ROS issue - RHIROS-1123
                                // TODO: Should be removed once fixed in kruize
```

internal/services/recommender.go
```diff
@@ -54,7 +54,13 @@ func ProcessEvent(msg *kafka.Message) {
...
+				if !(len(notifications) > 1 && notifications[0].Notificationtype == "info" && notifications[0].Message == "There is not enough data available to generate a recommendation.") {
```
kgaikwad, line 63:
```suggestion
				if !(len(notifications) > 1 && notifications[0].Type == "info" && notifications[0].Message == "There is not enough data available to generate a recommendation.") {
```

Looking at `notifications` object from [here](https://github.com/kruize/autotune/issues/770), it says `type` not `notification_type`. 
OR is it something related to golang convention?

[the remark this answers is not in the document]
patilsuraj767: We need `len(notifications) > 1` just to confirm notification array is not empty and we also need to verify the `Message`.  If needed we can skip the `Notificationtype` but I think it is good to have. 
Below is the exact json which we get. 
```
"notifications": [
                        {
                          "type": "info",
                          "message": "There is not enough data available to generate a recommendation."
                        }
                      ]

```

kgaikwad, line 63: Can we declare any variable on top and make use of it here so that may be easy to locate? Thoughts?

kgaikwad (changes requested):

internal/services/recommender.go
patilsuraj767: Yes, it is related to golang - [here](https://github.com/RedHatInsights/ros-ocp-backend/blob/main/internal/types/kruizePayload/common.go#L86)

patilsuraj767: Are you asking for variable for string -` There is not enough data available to generate a recommendation.`?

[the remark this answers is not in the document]
patilsuraj767: It should be `len(notifications) > 0`. I will make the changes.

kgaikwad:
Thanks @patilsuraj767 for addressing all comments.

PR #78, opened 2023-05-17
Point testing plugin in clowdapp

digitronik:
it will help us to debug the pr_check.sh integration test run. 

We need to tell clowder deployment to use this plugin so that we cand deploy the application and spin iqe-pod for further debugging.


patilsuraj767 (approved):
LGTM :+1:

PR #81, opened 2023-05-24
Fixes #RHIROS-1126 - Handle daily data from operator

patilsuraj767:


clowdapp.yaml
```diff
@@ -290,3 +296,5 @@ parameters:
   value: "1"
 - name: MACHINE_POOL_OPTION
   value: ''
+- name: LOG_LEVEL
+  value: 'INFO'
```
saltgen, line 300: @patilsuraj767 +1 for setting this here :ok_hand:

internal/services/report_processor.go
```diff
@@ -113,6 +114,14 @@ func ProcessReport(msg *kafka.Message) {
 				k8s_object_name,
 			)
 
+			if workload_metrics, err := model.GetWorkloadMetricsForTimestamp(experiment_name, interval_end); err != nil {
+				log.Errorf("Error while checking for workload_metrics record: %s", err)
+				continue
+			} else if !reflect.ValueOf(workload_metrics).IsZero() {
```
saltgen, line 120:
```suggestion
			} else if workload_metrics == (WorkloadMetrics{}) {
```
In case this works, we can avoid depending on `reflect`

patilsuraj767: I guess I tried that but it did not worked.

internal/services/report_processor.go
```diff
@@ -113,6 +114,14 @@ func ProcessReport(msg *kafka.Message) {
...
+				log.Debugf("worload_metrics table already has data for interval_end time: %v.", interval_end)
+				continue
```
kgaikwad, line 122: @patilsuraj767  - If metrics data at interval_end time is already available, do we still need to continue with further processing?

kgaikwad: Ohh, got it. If there are multiple slice data of 6hrs it should continue to process it.

saltgen (approved):
LGTM!

PR #82, opened 2023-05-25
RHIROS-1146 Report processor exits on Create failures

saltgen:
Changes:

Scenarios where-in code flow should not exit ~ `return` in case db record creation fails

### DIFF
--- a/internal/services/report_processor.go
+++ b/internal/services/report_processor.go
@@ -131,7 +131,7 @@ func ProcessReport(msg *kafka.Message) {
 			}
 			if err := workload.CreateWorkload(); err != nil {
 				log.Errorf("unable to get or add record to workloads table: %v. Error: %v", workload, err)
-				return
+				continue
 			}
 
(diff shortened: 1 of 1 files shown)


saltgen:
Need to check other `return`s and test locally, in-progress ...

saltgen:
Screenshot from 2023-05-25 17-19-57

Looks like we are good to go. 
@patilsuraj767  Q: Does below statement sound correct, w.r.t. count?
`Total records upload to rosocp.kruize.experiments = 3  service=rosocp`

patilsuraj767:
@saltgen if the csv which you submitted has 3 valid records with valid workload_type then output looks correct.

saltgen:
@patilsuraj767 I have used the ros-ocp-usage.csv from this repo, after updating the `owner_kind` and `workload_type` in the first entry.

patilsuraj767 (approved):
Thanks for the PR @saltgen LGTM 👍

PR #88, opened 2023-06-08
test kruize image via PR Check

saltgen:


patilsuraj767:
Nice hack :) This will solve kruize issue.

PR #89, opened 2023-06-09
Fixes #RHIROS-1185 - Fixed join query in workload_metrics model

patilsuraj767:


saltgen (approved):
LGTM :+1:

saltgen:
@patilsuraj767 The join looks fine but I think a simple get query should work here using experiment and end time filters.
If additionally, we need Workload record's data we can use gorm's preload feature.

patilsuraj767:
> @patilsuraj767 The join looks fine but I think a simple get query should work here using experiment and end time filters. If additionally, we need Workload record's data we can use gorm's preload feature.

Yes, we can optimize the query.

patilsuraj767:
Thanks for the review and ack @saltgen For now I will merge the PR.

PR #96, opened 2023-06-26
RHIROS-1134 Remove implicit rounding

saltgen:
TL;DR 493311537.55 bytes should transform to 470.45 MiB and not 470.46 MiB

There is a bit of rounding by the the fmt.Sprintf function, which is un-necessary atm.

Please don't quote me on the validity of the values :stuck_out_tongue:

### DIFF
--- a/internal/api/utils.go
+++ b/internal/api/utils.go
@@ -8,6 +8,7 @@ import (
 	"strings"
 	"time"
 	"encoding/json"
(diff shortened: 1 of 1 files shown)


patilsuraj767 (approved):

patilsuraj767:
Thanks @saltgen. 
Merging this PR.

PR #103, opened 2023-07-11
Fixes #RHIROS-1214 - Sources integration

patilsuraj767:
This PR includes a new micro service name - housekeeper. Current job of housekeeper service is to listen to the sources event kafka topic `platform.sources.event-stream` and look for the event_type `Application.destroy` based on the event housekeeper service deletes the OpenShift cluster and its associated workload metrics and recommendation from ROS DB.


internal/services/houseKeeper.go
```diff
@@ -0,0 +1,62 @@
...
+
+	"github.com/redhatinsights/ros-ocp-backend/internal/config"
+	"github.com/redhatinsights/ros-ocp-backend/internal/kafka"
+	"github.com/redhatinsights/ros-ocp-backend/internal/logging"
+	"github.com/redhatinsights/ros-ocp-backend/internal/model"
+	"github.com/redhatinsights/ros-ocp-backend/internal/types"
+	"github.com/redhatinsights/ros-ocp-backend/internal/utils/sources"
+)
+
+var cost_app_id int
```
saltgen, line 19: Would it be possible to use this value a function param instead of a global? I don't see much usage apart from sourcesListener.

saltgen (changes requested):

internal/services/houseKeeper.go
patilsuraj767: We cannot directly pass `cost_app_id` to the `sourcesListener`. If you see `StartHouseKeeperService()`function at the end it calls `kafka.StartConsumer()` and not `sourcesListener`. 
`kafka.StartConsumer()` has this signature - `StartConsumer(kafka_topic string, handler func(msg *kafka.Message))`. handler which in our case is `sourcesListener` should be a function which only has one argument of type - `*k.Message`

clowdapp.yaml
```diff
@@ -130,6 +131,30 @@ objects:
...
+            memory: ${MEMORY_LIMIT}
+        env:
+          - name: CLOWDER_ENABLED
+            value: ${CLOWDER_ENABLED}
+          - name: SSL_CERT_DIR
+            value: ${SSL_CERT_DIR}
+          - name: SERVICE_NAME
+            value: "rosocp-housekeeper"
+          - name: CW_LOG_STREAM_NAME
+            value: "rosocp-backend"
```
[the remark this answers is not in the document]
patilsuraj767: Done

internal/services/houseKeeper.go
patilsuraj767: `sourcesListener` is called every time when there is new event on kafka topic. If we call `GetCostApplicationID` inside the `sourcesListener` Then we will every time make an unnecessary API call to sources to get the cost_app_id. That is the reason I called `GetCostApplicationID` at the beginning when the service starts. 

Cost team also does the same - https://github.com/project-koku/koku/blob/main/koku/sources/kafka_listener.py#L388

saltgen: Ah I see, makes sense then

saltgen (approved):
Thanks @patilsuraj767 for the changes!

patilsuraj767:
Thanks for reviewing the PR @saltgen. Merging it.

PR #104, opened 2023-07-19
Fixes #RHIROS-1267 - Create separate table store historical recommend…

patilsuraj767:
…ation


migrations/000007_rename_recommendation_sets_table.up.sql
```diff
@@ -0,0 +1,7 @@
+ALTER TABLE recommendation_sets RENAME TO historical_recommendation_sets;
```
kgaikwad, line 1: Instead of altering the table, we can make this transition smooth for end users by creating table  as recommendation_sets and keep last reported records in recommendation_sets?

internal/services/recommender.go
```diff
@@ -75,6 +75,19 @@ func ProcessEvent(msg *kafka.Message) {
...
+
+				// Create entry into HistoricalRecommendationSet table.
+				historicalRecommendationSet := model.HistoricalRecommendationSet{
+					WorkloadID:          kafkaMsg.WorkloadID,
+					ContainerName:       container.Container_name,
+					MonitoringStartTime: v.Duration_based.Short_term.Monitoring_start_time,
+					MonitoringEndTime:   v.Duration_based.Short_term.Monitoring_end_time,
+					Recommendations:     marshalData,
+				}
+				if err := historicalRecommendationSet.CreateHistoricalRecommendationSet(); err != nil {
```
kgaikwad, line 87: Here, it creates entry in both tables but not deleting an existing entry from recommendation_set table. Is it intentional?

kgaikwad (changes requested):
Two inline comments added.

kgaikwad:
Nitpick: 
In addition to this change, at this [line-73](https://github.com/RedHatInsights/ros-ocp-backend/pull/104/files#diff-b4b37b1e14f4e2fb303da4d9adef5daebfbd6bc938a2488f78975b656e71b89aR73) it says `"unable to get or add record to recommendation set table: %v. Error: %v"`. Can we simply make it -  "Unable to save a record into recommendation set: %v. Error: %v" ?

internal/services/recommender.go
```diff
@@ -75,6 +75,19 @@ func ProcessEvent(msg *kafka.Message) {
 				} else {
 					log.Infof("Recommendation saved for experiment - %s and end_interval - %s", kafkaMsg.Experiment_name, recommendationSet.MonitoringEndTime)
 				}
+
+				// Create entry into HistoricalRecommendationSet table.
+				historicalRecommendationSet := model.HistoricalRecommendationSet{
```
kgaikwad, line 80: In my opinion, no one really needs to care about this internal implementation as it is just creating copy of record into another table.
So, Will it make sense to extract out creation of recommendation set and historical set into a method so that it can be reuse to other places?

migrations/000007_rename_recommendation_sets_table.up.sql
patilsuraj767: recommendation_sets table is already created and it has multiple recommendation for same workload so it is hard to create new historical_recommendation_sets table and move the older recommendation to it. Instead I thought to just rename the recommendation_sets to historical_recommendation_sets which contains the multiple recommendations and a new recommendation_sets is created were only latest 1 recommendation per workload will be saved.

internal/services/recommender.go
patilsuraj767: In the PR - [here](https://github.com/RedHatInsights/ros-ocp-backend/pull/104/files#diff-fe0fff74011513527c16ac9434707cd2add9039b21e9967b12f487f42832be49R86) we have db clause OnConflict. so if we try to add new entry in the table and if table already has entry for that workload then it just updates the recommendation in that entry. Hence no need of deletion.

internal/services/recommender.go
patilsuraj767: To extract this logic to some other function we will need to export few internal type which we have declared under `/internal/type` this will create bit much refactoring of code. Can we do this later in some other PR?

migrations/000009_modify_recommendation_sets_table.up.sql
```diff
@@ -0,0 +1,3 @@
+ALTER TABLE recommendation_sets DROP CONSTRAINT IF EXISTS UQ_Recommendation;
+ALTER TABLE recommendation_sets
+ADD CONSTRAINT UQ_Recommendation UNIQUE (workload_id, container_name);
```
kgaikwad, line 3:
```suggestion
ALTER TABLE recommendation_sets 
DROP CONSTRAINT IF EXISTS UQ_Recommendation,
ADD CONSTRAINT UQ_Recommendation UNIQUE (workload_id, container_name);
```

migrations/000009_modify_recommendation_sets_table.down.sql
```diff
@@ -0,0 +1,3 @@
+ALTER TABLE recommendation_sets DROP CONSTRAINT IF EXISTS UQ_Recommendation;
+ALTER TABLE recommendation_sets
+ADD CONSTRAINT UQ_Recommendation UNIQUE (workload_id, container_name, monitoring_end_time);
```
kgaikwad, line 3:
```suggestion
ALTER TABLE recommendation_sets 
DROP CONSTRAINT IF EXISTS UQ_Recommendation,
ADD CONSTRAINT UQ_Recommendation UNIQUE (workload_id, container_name, monitoring_end_time);
```

migrations/000008_keep_last_reported.up.sql
```diff
@@ -0,0 +1,4 @@
+DELETE FROM recommendation_sets WHERE id IN 
+(SELECT id FROM recommendation_sets r1 JOIN 
+(SELECT workload_id, container_name, MAX(monitoring_end_time) AS monitoring_end_time FROM recommendation_sets GROUP BY workload_id, container_name) AS r2 
+ON r1.workload_id = r2.workload_id AND r1.container_name = r2.container_name AND r1.monitoring_end_time != r2.monitoring_end_time);
```
kgaikwad, line 4:
```suggestion
DELETE FROM recommendation_sets WHERE id NOT IN 
(SELECT id FROM recommendation_sets r1 JOIN 
(SELECT workload_id, container_name, MAX(monitoring_end_time) AS monitoring_end_time FROM recommendation_sets GROUP BY workload_id, container_name) AS r2 
ON r1.workload_id = r2.workload_id AND r1.container_name = r2.container_name AND r1.monitoring_end_time == r2.monitoring_end_time);
```
Will it be good idea to have another way around?

kgaikwad (changes requested):

kgaikwad (approved):

kgaikwad:
Thank you @patilsuraj767 for addressing all comments patiently :fireworks:

PR #106, opened 2023-07-21
Fixes #RHIROS-1218 - Get recommendation only for the last data point

patilsuraj767:


internal/services/report_processor.go
```diff
@@ -75,122 +75,138 @@ func ProcessReport(msg *kafka.Message) {
...
-				WorkloadType:    workload.WorkloadType(k8s_object_type),
-				WorkloadName:    k8s_object_name,
-				Containers:      container_names,
-				MetricsUploadAt: time.Now(),
-			}
-			if err := workload.CreateWorkload(); err != nil {
-				log.Errorf("unable to get or add record to workloads table: %v. Error: %v", workload, err)
-				continue
-			}
+			sort.SliceStable(keys, func(i, j int) bool {
```
saltgen, line 95:
```suggestion
			sort.SlicesTable(keys, func(i, j int) bool {
```
@patilsuraj767 Could you add a comment here, especially regarding the purpose of `Before`?

internal/services/report_processor.go
```diff
@@ -75,122 +75,138 @@ func ProcessReport(msg *kafka.Message) {
...
+					namespace,
+					k8s_object_type,
+					k8s_object_name,
+				)
+
+				if workload_metrics, err := model.GetWorkloadMetricsForTimestamp(experiment_name, interval_end); err != nil {
+					log.Errorf("Error while checking for workload_metrics record: %s", err)
+					continue
+				} else if !reflect.ValueOf(workload_metrics).IsZero() {
+					log.Debugf("worload_metrics table already has data for interval_end time: %v.", interval_end)
```
saltgen, line 131:
```suggestion
					log.Debugf("workload_metrics table already has data for interval_end time: %v.", interval_end)
```

internal/services/report_processor.go
```diff
@@ -75,122 +75,138 @@ func ProcessReport(msg *kafka.Message) {
...
+					Namespace:       namespace,
+					WorkloadType:    workload.WorkloadType(k8s_object_type),
+					WorkloadName:    k8s_object_name,
+					Containers:      container_names,
+					MetricsUploadAt: time.Now(),
 				}
-				if err := workload_metric.CreateWorkloadMetrics(); err != nil {
-					log.Errorf("unable to add record to workload_metrics table: %v. Error: %v", workload_metric, err)
+				if err := workload.CreateWorkload(); err != nil {
+					log.Errorf("unable to get or add record to workloads table: %v. Error: %v", workload, err)
```
saltgen, line 152: @patilsuraj767 Does workload.CreateWorkload have a native get or create functionality?

internal/services/report_processor.go
```diff
@@ -75,122 +75,138 @@ func ProcessReport(msg *kafka.Message) {
...
-				Fetch_time:          time.Now().UTC().Add(time.Second * time.Duration(waittime)),
-				Monitoring_end_time: interval_end.String(),
-				K8s_object:          k8s_object,
-				Attempt:             1,
-				Kafka_request_msg:   kafkaMsg,
-			}
+				usage_data_byte, err := kruize.Update_results(experiment_name, k8s_object)
+				if err != nil {
+					log.Error(err)
+					continue
```
saltgen, line 159: @patilsuraj767 Should we attempt this continue post some secs of sleep, considering API requests fail?

saltgen (changes requested):

internal/services/report_processor.go
patilsuraj767: `SliceStable` is a function in `sort` package we cannot change the function name.  Using `sort.SliceStable` we are basically sorting the group of deployment into chronological order based on `interval_end` time

internal/services/report_processor.go
patilsuraj767: It has create or update functionality, while inserting if it is found that workload is already present then it updates the `containers` and `metrics_upload_at` field of the same record. - https://github.com/RedHatInsights/ros-ocp-backend/blob/main/internal/model/workload.go#L36

internal/services/report_processor.go
patilsuraj767: We cannot add sleep in `if block` because API call has been finished until code reaches to `if block`. In our code base everything work synchronously(http call is not async) so I don't see benefit in adding sleep.

internal/services/report_processor.go
saltgen: Okay, could you please modify the comment to `create or update`?

internal/services/report_processor.go
saltgen: IIUC, the flow reaches `continue` only if there is an error, shall we add some code that checks if the API url responds and then does a `continue`?

patilsuraj767: Actual http request does not happens from here.  `kruize.Update_results` is a custom function defined - [here](https://github.com/RedHatInsights/ros-ocp-backend/blob/main/internal/utils/kruize/kruize_api.go#L91) If have added a response check on [line 114](https://github.com/RedHatInsights/ros-ocp-backend/blob/main/internal/utils/kruize/kruize_api.go#L114).

internal/services/report_processor.go
kgaikwad:
```suggestion
					log.Errorf("unable to save workload record: %v. Error: %v", workload, err)
```
How about this?

patilsuraj767: Done

saltgen (approved):
LGTM, thanks @patilsuraj767!

kgaikwad (approved):
Codewise changes look good. 
It seems little inconsistency with logs but can be addressed separately.

patilsuraj767:
Thanks for the review @saltgen and @kgaikwad 
Merging this PR.

PR #114, opened 2023-08-02
Changing SOURCES_API_PREFIX

patilsuraj767:
In local setup API v3.1 had worked but looking at the COST code - [here](https://github.com/project-koku/koku/blob/main/koku/sources/config.py#L26). API version should be `v1.0`

### DIFF
--- a/internal/config/config.go
+++ b/internal/config/config.go
@@ -158,7 +158,7 @@ func initConfig() {
 
 	}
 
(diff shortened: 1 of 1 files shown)


patilsuraj767:
Ci failing with below issue - 
```
 ERROR: deploy failed: component 'sed-frontend' is not unique, found in apps: ['config-manager', 'insights-ephemeral']
```
This is a known issue and is posted in internal slack channel.

PR #122, opened 2023-09-07
RHIROS-1327 - Adding support for report having more than 24hrs o…

patilsuraj767:
…f data.

### DIFF
--- a/internal/config/config.go
+++ b/internal/config/config.go
@@ -31,8 +31,9 @@ type Config struct {
 	KafkaCA               string
 
 	// Kruize config
-	KruizeUrl      string `mapstructure:"KRUIZE_URL"`
-	KruizeWaitTime string `mapstructure:"KRUIZE_WAIT_TIME"`
+	KruizeUrl                string `mapstructure:"KRUIZE_URL"`
+	KruizeWaitTime           string `mapstructure:"KRUIZE_WAIT_TIME"`
+	KruizeMaxBulkUpdateLimit int    `mapstructure:"KRUIZE_MAX_BULK_UPDATE_LIMIT"`
 
 	// Database config
 	DBName     string
@@ -162,6 +163,7 @@ func initConfig() {
 	viper.SetDefault("SERVICE_NAME", "rosocp")
 	viper.SetDefault("API_PORT", "8000")
 	viper.SetDefault("KRUIZE_WAIT_TIME", "30")
+	viper.SetDefault("KRUIZE_MAX_BULK_UPDATE_LIMIT", 100)
 	viper.SetDefault("KAFKA_CONSUMER_GROUP_ID", "ros-ocp")
 	viper.SetDefault("KAFKA_AUTO_COMMIT", true)
 	viper.SetDefault("LOG_LEVEL", "INFO")
--- a/internal/config/config.go
+++ b/internal/config/config.go
@@ -31,9 +31,9 @@ type Config struct {
 	KafkaCA               string
 
 	// Kruize config
-	KruizeUrl                string `mapstructure:"KRUIZE_URL"`
-	KruizeWaitTime           string `mapstructure:"KRUIZE_WAIT_TIME"`
-	KruizeMaxBulkUpdateLimit int    `mapstructure:"KRUIZE_MAX_BULK_UPDATE_LIMIT"`
+	KruizeUrl              string `mapstructure:"KRUIZE_URL"`
+	KruizeWaitTime         string `mapstructure:"KRUIZE_WAIT_TIME"`
+	KruizeMaxBulkChunkSize int    `mapstructure:"KRUIZE_MAX_BULK_CHUNK_SIZE"`
 
 	// Database config
 	DBName     string
@@ -163,7 +163,7 @@ func initConfig() {
 	viper.SetDefault("SERVICE_NAME", "rosocp")
 	viper.SetDefault("API_PORT", "8000")
 	viper.SetDefault("KRUIZE_WAIT_TIME", "30")
-	viper.SetDefault("KRUIZE_MAX_BULK_UPDATE_LIMIT", 100)
+	viper.SetDefault("KRUIZE_MAX_BULK_CHUNK_SIZE", 100)
 	viper.SetDefault("KAFKA_CONSUMER_GROUP_ID", "ros-ocp")
 	viper.SetDefault("KAFKA_AUTO_COMMIT", true)
 	viper.SetDefault("LOG_LEVEL", "INFO")
--- a/internal/services/report_processor.go
+++ b/internal/services/report_processor.go
@@ -10,7 +10,6 @@ import (
 	"github.com/go-gota/gota/dataframe"
 	"github.com/go-playground/validator/v10"
 
-	"github.com/redhatinsights/ros-ocp-backend/internal/config"
 	"github.com/redhatinsights/ros-ocp-backend/internal/logging"
 	"github.com/redhatinsights/ros-ocp-backend/internal/model"
 	"github.com/redhatinsights/ros-ocp-backend/internal/types"
@@ -21,7 +20,6 @@ import (
 
 func ProcessReport(msg *kafka.Message) {
 	log := logging.GetLogger()
-	cfg := config.GetConfig()
 	validate := validator.New()
 	var kafkaMsg types.KafkaMsg
 	if !json.Valid([]byte(msg.Value)) {
@@ -119,7 +117,7 @@ func ProcessReport(msg *kafka.Message) {
 				continue
 			}
 
-			k8s_object_chunks := utils.ChunkK8sobjectSlice(k8s_object, cfg.KruizeMaxBulkUpdateLimit)
+			k8s_object_chunks := utils.SliceK8sObjectToChunks(k8s_object)
 
 			for _, chunk := range k8s_object_chunks {
 				usage_data_byte, err := kruize.Update_results(experiment_name, chunk)
--- a/internal/services/report_processor.go
+++ b/internal/services/report_processor.go
@@ -122,14 +122,22 @@ func ProcessReport(msg *kafka.Message) {
 			for _, chunk := range k8s_object_chunks {
 				usage_data_byte, err := kruize.Update_results(experiment_name, chunk)
 				if err != nil {
-					log.Error(err)
+					log.Error(err, experiment_name)
 					continue
 				}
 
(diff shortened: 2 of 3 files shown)


kgaikwad:
@patilsuraj767  - could you update ticket ID  to [RHIROS-1327](https://issues.redhat.com/browse/RHIROS-1327) in commit message as well as title.

internal/utils/utils.go
```diff
@@ -126,6 +126,14 @@ func ConvertStringToTime(data string) (time.Time, error) {
 
 }
 
+func ConvertISO8601StringToTime(data string) (time.Time, error) {
+	dateTime, err := time.Parse("2006-01-02T15:04:05.000Z", data)
+	if err != nil {
+		return time.Time{}, fmt.Errorf("unable to convert string to time: %s", err)
+	}
+	return dateTime, nil
```
kgaikwad, line 134: Here, this method returns two values i.e dateTime and `nil`.  Why Can't we return just dataTime value as second value will be `nil` always?

kgaikwad (changes requested):
Added few inline suggestion.

internal/services/report_processor.go
```diff
@@ -117,44 +119,48 @@ func ProcessReport(msg *kafka.Message) {
...
-			usage_data_byte, err := kruize.Update_results(experiment_name, k8s_object)
-			if err != nil {
-				log.Error(err)
-				continue
-			}
-
-			for _, data := range usage_data_byte {
+			k8s_object_chunks := utils.ChunkK8sobjectSlice(k8s_object, cfg.KruizeMaxBulkUpdateLimit)
 
-				interval_start_time, _ := utils.ConvertStringToTime(data.Interval_start_time)
```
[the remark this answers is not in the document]
patilsuraj767: It is been used - [here](https://github.com/RedHatInsights/ros-ocp-backend/blob/main/internal/utils/utils.go#L132)

internal/utils/utils.go
patilsuraj767: It is not returning nil always, If you check line 132 we are returning error on it.

internal/config/config.go
```diff
@@ -31,8 +31,9 @@ type Config struct {
 	KafkaCA               string
 
 	// Kruize config
-	KruizeUrl      string `mapstructure:"KRUIZE_URL"`
-	KruizeWaitTime string `mapstructure:"KRUIZE_WAIT_TIME"`
+	KruizeUrl                string `mapstructure:"KRUIZE_URL"`
+	KruizeWaitTime           string `mapstructure:"KRUIZE_WAIT_TIME"`
+	KruizeMaxBulkUpdateLimit int    `mapstructure:"KRUIZE_MAX_BULK_UPDATE_LIMIT"`
```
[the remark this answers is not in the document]
patilsuraj767: Resolved

internal/config/config.go
```diff
@@ -162,6 +163,7 @@ func initConfig() {
 	viper.SetDefault("SERVICE_NAME", "rosocp")
 	viper.SetDefault("API_PORT", "8000")
 	viper.SetDefault("KRUIZE_WAIT_TIME", "30")
+	viper.SetDefault("KRUIZE_MAX_BULK_UPDATE_LIMIT", 100)
```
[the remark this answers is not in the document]
patilsuraj767: Resolved

internal/services/report_processor.go
```diff
@@ -117,44 +119,48 @@ func ProcessReport(msg *kafka.Message) {
...
 			}
 
-			usage_data_byte, err := kruize.Update_results(experiment_name, k8s_object)
-			if err != nil {
-				log.Error(err)
-				continue
-			}
-
-			for _, data := range usage_data_byte {
+			k8s_object_chunks := utils.ChunkK8sobjectSlice(k8s_object, cfg.KruizeMaxBulkUpdateLimit)
```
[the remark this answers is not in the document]
patilsuraj767: Resolved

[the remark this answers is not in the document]
patilsuraj767: Resolved

internal/utils/utils.go
kgaikwad: Yes , it is returning error on that line. But at the end, you used a blank identifier (i.e `_` ) to capture second value at method calls. That means, it is not required, correct?

Will it be possible to return just value and see if we can log here directly an error? OR may be handle scenario what if there is an error?

internal/services/report_processor.go
```diff
@@ -117,44 +117,48 @@ func ProcessReport(msg *kafka.Message) {
...
-
-				if workload_metrics, err := model.GetWorkloadMetricsForTimestamp(experiment_name, interval_end_time); err != nil {
-					log.Errorf("Error while checking for workload_metrics record: %s", err)
-					continue
-				} else if !reflect.ValueOf(workload_metrics).IsZero() {
-					log.Debugf("workload_metrics table already has data for interval_end time: %v.", interval_end_time)
+			for _, chunk := range k8s_object_chunks {
+				usage_data_byte, err := kruize.Update_results(experiment_name, chunk)
+				if err != nil {
+					log.Error(err)
```
kgaikwad, line 125: Could you confirm what details present under `err`? Will it be sufficient for debugging?


If something went wrong/failure occurs for `update_results` api with certain chunk, there will be data loss. Am I correct? 
If this is so, is it expected behavior?

kgaikwad (changes requested):

internal/services/report_processor.go
```diff
@@ -117,44 +117,48 @@ func ProcessReport(msg *kafka.Message) {
...
 			}
 
-			usage_data_byte, err := kruize.Update_results(experiment_name, k8s_object)
-			if err != nil {
-				log.Error(err)
-				continue
-			}
-
-			for _, data := range usage_data_byte {
+			k8s_object_chunks := utils.SliceK8sObjectToChunks(k8s_object)
```
saltgen, line 120: IMO it would be logically straight-forward to compare `len(k8s_objects)` and `cfg.KruizeMaxBulkChunkSize` before using this util function, such that we use it only if we need to.

saltgen (changes requested):

internal/services/report_processor.go
kgaikwad: Yes, along with unique identifier for resource, it is good to include chunk specific info. example -  time duration

internal/services/report_processor.go
patilsuraj767: No we cannot because `k8s_object` is of type `[]map[string]interface{}` and  `k8s_object_chunks` is of type `[][]map[string]interface{}`.

internal/services/report_processor.go
patilsuraj767: Yes @kgaikwad if something went wrong/failure occurs for update_results api with certain chunk, there will be data loss but we cannot help it.  example - if suppose kruize application is down then update_results api will fail and we cannot do anything here to stop further API requests.

patilsuraj767: Yes are not dividing chunks based on time stamp, means it is not the case that every metrics in the chunk have the same time stamp. So I think timestamp cannot be the unique identifier.

kgaikwad: >  Yes @kgaikwad if something went wrong/failure occurs for update_results api with certain chunk, there will be data loss but we cannot help it. example - if suppose kruize application is down then update_results api will fail and we cannot do anything here to stop further API requests.

In my opinion, we should think on such scenarios and come with handling to minimize data loss. May be not for this PR but it is better to do brainstorming around.

> So I think timestamp cannot be the unique identifier.

Then could you please include experiment name and any field info that might be specific to that request? So it will be easy for identifying that this upload request not processed completely, data loss observed.

patilsuraj767: I have added the `experiment_name` in the log and by default `request_id` will be getting logged with the every log message.

internal/services/report_processor.go
```diff
@@ -117,44 +117,56 @@ func ProcessReport(msg *kafka.Message) {
...
 				}
 
-				for _, container := range data.Kubernetes_objects[0].Containers {
-					container_usage_metrics, err := json.Marshal(container.Metrics)
+				for _, data := range usage_data_byte {
+
+					interval_start_time, err := utils.ConvertISO8601StringToTime(data.Interval_start_time)
 					if err != nil {
-						log.Errorf("Unable to marshal container usage data: %v", err)
+						log.Errorf("Unable to convert start time ISO8601 string to time: %s", err)
```
kgaikwad, line 133:
```suggestion
						log.Errorf("Error for start time: %s", err)
```
As `err` object already includes the error message, no need of full message here.  To identify which field, you can include field name here.

internal/services/report_processor.go
```diff
@@ -117,44 +117,56 @@ func ProcessReport(msg *kafka.Message) {
...
-
-					workload_metric := model.WorkloadMetrics{
-						WorkloadID:    workload.ID,
-						ContainerName: container.Container_name,
-						IntervalStart: interval_start_time,
-						IntervalEnd:   interval_end_time,
-						UsageMetrics:  container_usage_metrics,
+					interval_end_time, err := utils.ConvertISO8601StringToTime(data.Interval_end_time)
+					if err != nil {
+						log.Errorf("Unable to convert end time ISO8601 string to time: %s", err)
```
kgaikwad, line 138: ditto
```suggestion
						log.Errorf("Error for end time: %s", err)
```

kgaikwad (changes requested):

internal/services/report_processor.go
kgaikwad: For first part, could you add a backlog ticket so that it won't get ignored by time. Thanks!

internal/utils/utils.go
kgaikwad: @patilsuraj767, captured and added log outside the method.

kgaikwad (approved):
Not tested but codewise looks good to me.

internal/services/report_processor.go
saltgen: If we use the util function only if len(k8s_objects) > 100 then I think we could set k8s_object as [][]map[string]interface{} on another var for easy looping

patilsuraj767: Done, Made the changes.

saltgen (approved):
LGTM! Thanks @patilsuraj767!

patilsuraj767:
Thank you @saltgen and @kgaikwad for the review. 
Merging the PR.

PR #126, opened 2023-09-15
RHIROS-1345 Probes for processor

saltgen:


clowdapp.yaml
```diff
@@ -28,6 +28,26 @@ objects:
           limits:
             cpu: ${CPU_LIMIT}
             memory: ${MEMORY_LIMIT}
+        livenessProbe:
+          failureThreshold: 3
+          httpGet:
+            path: /
```
patilsuraj767, line 34: Here it should be `/metrics` and not `/`. Ref - https://github.com/RedHatInsights/ros-ocp-backend/blob/main/internal/utils/utils.go#L183

patilsuraj767 (changes requested):

kgaikwad:
\retest

clowdapp.yaml
saltgen: Thanks @patilsuraj767, change pushed, do the the config params look fine?

patilsuraj767 (approved):
LGTM 👍

clowdapp.yaml
patilsuraj767: Config look good for now, Later if we think that there are many http requests coming to the pod we can increase the `periodSeconds` time.

PR #131, opened 2023-09-25
Fixes: #RHIROS-1365 - Removed unnecessary workload_metrics check

patilsuraj767:
`workload_metrics` check on line 149 was initially written to optimize the codebase, that is to avoid the `for loop` on line 157 in case if duplicate report is uploaded. But in production it is very rare that duplicate report will arrive because cost operator will never do that. 
This check is hampering the DB performance hence removing it.

Screenshot 2023-09-25 at 9 45 07 PM


saltgen (approved):
LGTM :+1:  @kgaikwad Do we need a card for this one?

PR #136, opened 2023-10-12
Added debug logs

patilsuraj767:


internal/utils/kruize/kruize_api.go
```diff
@@ -102,15 +102,16 @@ func Update_results(experiment_name string, k8s_object []map[string]interface{})
 
 	// Update metrics to kruize experiment
 	url := cfg.KruizeUrl + "/updateResults"
-
+	log.Debugf("\n Sending /updateResult request to kruize with payload - %s \n", string(postBody))
```
kgaikwad, line 105: Just a thought - For better log filtering, do you think we should include `experiment_name` / `request_id` / `org_id` into these log statements.

kgaikwad (changes requested):
Just added a small inline suggestion, please ignore if not required.

internal/utils/kruize/kruize_api.go
patilsuraj767: `experiment_name` and `org_id` will be present in the `postBody` and we do not have request_id available in this part of the code.

PR #139, opened 2023-10-19
RHIROS-1341 Fix sorting for workload type

srdeotarse:
Added new migration to alter ``workloadtype`` type of column ``workloads.workload_type`` to correct order of constants in ``workloadtype`` enum

Also removed ``make local-upload-data`` and updated ``make get-recommendations`` as suggested by @patilsuraj767

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -105,12 +105,6 @@ else
 	@ echo "Env not defined"
 endif
 
-local-upload-data:
-	curl -vvvv -F "upload=@$(file);type=application/application/vnd.redhat.hccm.tar+tgz" \
-		-H "x-rh-identity: ${b64_identity}" \
-		-H "x-rh-request_id: testtesttest" \
-		http://localhost:${INGRESS_PORT}/api/ingress/v1/upload
-
 upload-msg-to-rosocp:
 	echo ${ros_ocp_msg} | docker-compose -f scripts/docker-compose.yml exec -T kafka kafka-console-producer --topic hccm.ros.events  --broker-list localhost:29092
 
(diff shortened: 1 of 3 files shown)


Makefile
```diff
@@ -105,22 +105,16 @@ else
...
-
 upload-msg-to-rosocp:
 	echo ${ros_ocp_msg} | docker-compose -f scripts/docker-compose.yml exec -T kafka kafka-console-producer --topic hccm.ros.events  --broker-list localhost:29092
 
 
 get-recommendations:
 ifdef env
 	$(eval APIPOD=$(shell oc get pods -o custom-columns=POD:.metadata.name --no-headers -n ${env} | grep ros-ocp-backend-api))
-	oc exec ${APIPOD} -c ros-ocp-backend-api -n ${env} -- /bin/bash -c 'curl -s -H "X-Rh-Identity: ${b64_identity}" -H "x-rh-request_id: testtesttest" http://localhost:8000/api/cost-management/v1/recommendations/openshift' | python -m json.tool
+	oc exec ${APIPOD} -c ros-ocp-backend-api -n ${env} -- /bin/bash -c 'curl -s -H "X-Rh-Identity: ${b64_identity}" -H "x-rh-request_id: testtesttest" http://localhost:8000/api/cost-management/v1/recommendations/openshift?start_date=2023-04-01' | python -m json.tool
```
upadhyeammit, line 115: Is it possible to not hardcode the start_date? Its my assumption that make file should have a way to pass cmd line arguments and then those can be used?

something like:
~~~
make get-recommendations start_date=2023-08-01
~~~

Makefile
```diff
@@ -105,22 +105,16 @@ else
...
 else
 	curl -s -H "x-rh-identity: ${b64_identity}" \
 		 -H "x-rh-request_id: testtesttest" \
-		 http://localhost:8000/api/cost-management/v1/recommendations/openshift | python -m json.tool
+		 http://localhost:8000/api/cost-management/v1/recommendations/openshift?start_date=2023-04-01 | python -m json.tool
```
upadhyeammit, line 119: Here as well if we can avoid the hardcoding of date?

migrations/000011_alter_workloads_table_workloadtype.up.sql
```diff
@@ -0,0 +1,9 @@
+ALTER TYPE workloadtype RENAME TO workloadtype_old;
```
upadhyeammit, line 1: I was under impression that if we rename the ENUM then table won't be able to reference it or we need to manually change the reference however I can see its updating the name in table as well. So from migration point of view it looks good to me!

srdeotarse: I have followed this [article](https://dev.to/yogski/dealing-with-enum-type-in-postgresql-1j3g) for reference.

Makefile
srdeotarse: The [ros-ocp-usage.csv](https://github.com/RedHatInsights/ros-ocp-backend/blob/main/scripts/samples/ros-ocp-usage.csv) file we are using to upload data to local kafka topic for processor to upload results to recommendations populated after this date. So, I had faced an issue as a newcomer that I was not able to fetch recommendations because I didn't knew what start date was.

Thanks @upadhyeammit for pointing this out. Maybe @patilsuraj767 has something to add to this?

saltgen:
@srdeotarse Any luck with the PR check failure?

srdeotarse:
Closing this PR as opened new PR - #140 from forked repo

PR #140, opened 2023-10-23
RHIROS-1341 fix sorting for workload type

srdeotarse:
Added new migration to alter ``workloadtype`` type of column ``workloads.workload_type`` to correct order of constants in ``workloadtype enum``

Also removed ``make local-upload-data`` and updated ``make get-recommendations`` as suggested by @patilsuraj767

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -38,6 +38,8 @@ endif
 .PHONY: golangci-lint
 GOLANGCILINT := $(LOCALBIN)/golangci-lint
 GOLANGCI_URL := https://raw.githubusercontent.com/golangci/golangci-lint/master/install.sh
+start_date := 1970-01-01
+
 golangci-lint: $(LOCALBIN)
 ifeq (,$(wildcard $(GOLANGCILINT)))
 	@ echo "📥 Downloading golangci-lint"
@@ -105,12 +107,6 @@ else
 	@ echo "Env not defined"
 endif
 
-local-upload-data:
-	curl -vvvv -F "upload=@$(file);type=application/application/vnd.redhat.hccm.tar+tgz" \
-		-H "x-rh-identity: ${b64_identity}" \
-		-H "x-rh-request_id: testtesttest" \
-		http://localhost:${INGRESS_PORT}/api/ingress/v1/upload
-
 upload-msg-to-rosocp:
 	echo ${ros_ocp_msg} | docker-compose -f scripts/docker-compose.yml exec -T kafka kafka-console-producer --topic hccm.ros.events  --broker-list localhost:29092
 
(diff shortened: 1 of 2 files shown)


Makefile
```diff
@@ -105,22 +105,16 @@ else
...
 get-recommendations:
 ifdef env
 	$(eval APIPOD=$(shell oc get pods -o custom-columns=POD:.metadata.name --no-headers -n ${env} | grep ros-ocp-backend-api))
-	oc exec ${APIPOD} -c ros-ocp-backend-api -n ${env} -- /bin/bash -c 'curl -s -H "X-Rh-Identity: ${b64_identity}" -H "x-rh-request_id: testtesttest" http://localhost:8000/api/cost-management/v1/recommendations/openshift' | python -m json.tool
+	oc exec ${APIPOD} -c ros-ocp-backend-api -n ${env} -- /bin/bash -c 'curl -s -H "X-Rh-Identity: ${b64_identity}" -H "x-rh-request_id: testtesttest" http://localhost:8000/api/cost-management/v1/recommendations/openshift?start_date=2023-04-01' | python -m json.tool
 else
 	curl -s -H "x-rh-identity: ${b64_identity}" \
 		 -H "x-rh-request_id: testtesttest" \
-		 http://localhost:8000/api/cost-management/v1/recommendations/openshift | python -m json.tool
+		 http://localhost:8000/api/cost-management/v1/recommendations/openshift?start_date=2023-04-01 | python -m json.tool
```
saltgen, line 119: @upadhyeammit Shall we consider Epoch date i.e. January 1, 1970. It is used as the starting reference for timekeeping in Unix and POSIX systems?
This would also let us avoid the additional step of mentioning a date on the make command

saltgen:
I'll test these changes with a stage data dump and revert shortly.

migrations/000011_alter_workloads_table_workloadtype.down.sql
```diff
@@ -0,0 +1,9 @@
+ALTER TYPE workloadtype RENAME TO workloadtype_old;
+
+CREATE TYPE workloadtype AS ENUM ('deployment', 'deploymentconfig', 'replicaset', 'replicationcontroller', 'statefulset', 'daemonset');
```
saltgen, line 3: Noticed the same order of values, [here](https://github.com/RedHatInsights/ros-ocp-backend/blob/main/migrations/000003_create_workloads_table.up.sql#L1). I don't think this will lead to issue resolution.

saltgen (changes requested):

migrations/000011_alter_workloads_table_workloadtype.up.sql
```diff
@@ -0,0 +1,9 @@
+ALTER TYPE workloadtype RENAME TO workloadtype_old;
+
+CREATE TYPE workloadtype AS ENUM ('daemonset', 'deployment', 'deploymentconfig', 'replicaset', 'replicationcontroller', 'statefulset');
```
srdeotarse, line 3: @saltgen Any custom ``TYPE`` enum value can't be deleted but only be appended. We will have to create new ``TYPE`` with proper ordering, rename previous ``TYPE `` to something different name then link new created TYPE to referenced tables ie ``workloads`` table in this case.

migrations/000011_alter_workloads_table_workloadtype.down.sql
saltgen: As discussed, change is now introduced in existing migration file.

migrations/000011_alter_workloads_table_workloadtype.up.sql
saltgen: Ditto

saltgen (approved):
Test with stage data dump, looks good.
Thanks @srdeotarse :+1:

saltgen:
@upadhyeammit Do you have any further comments?

upadhyeammit (approved):
The `start_date` param and its implementation looks good to me!

PR #144, opened 2023-10-30
Fixes #RHIROS-1401 - Dropping csv records with missing resource usage…

patilsuraj767:
… metrics

### DIFF
--- a/internal/utils/aggregator.go
+++ b/internal/utils/aggregator.go
@@ -5,11 +5,12 @@ import (
 
 	"github.com/go-gota/gota/dataframe"
 	"github.com/go-gota/gota/series"
+	"github.com/sirupsen/logrus"
 
 	w "github.com/redhatinsights/ros-ocp-backend/internal/types/workload"
 )
 
-func Aggregate_data(df dataframe.DataFrame) dataframe.DataFrame {
+func Aggregate_data(df dataframe.DataFrame, log *logrus.Entry) dataframe.DataFrame {
 	df = df.FilterAggregation(
 		dataframe.And,
 		dataframe.F{Colname: "owner_kind", Comparator: series.Neq, Comparando: ""},
@@ -40,6 +41,33 @@ func Aggregate_data(df dataframe.DataFrame) dataframe.DataFrame {
 
 	df = df.Mutate(s.Col("X0")).Rename("k8s_object_type", "X0")
 	df = df.Mutate(s.Col("X1")).Rename("k8s_object_name", "X1")
+
+	// filter out only valid workload type
+	df = df.Filter(
+		dataframe.F{
+			Colname:    "k8s_object_type",
+			Comparator: series.In,
+			Comparando: []string{
+				w.Daemonset.String(),
+				w.Deployment.String(),
+				w.Deploymentconfig.String(),
+				w.Replicaset.String(),
+				w.Replicationcontroller.String(),
+				w.Statefulset.String(),
+			}},
+	)
+
+	// Validation to check if metrics for cpuUsage, memoryUsage and memoryRSS are missing
+	df, no_of_dropped_records := filter_valid_csv_records(df)
+	if no_of_dropped_records != 0 {
+		invalidDataPoints.Add(float64(no_of_dropped_records))
+		log.Infof("Invalid records in CSV - %v", no_of_dropped_records)
+	}
+
+	if df.Nrow() == 0 {
+		return df
+	}
+
 	dfGroups := df.GroupBy(
 		"namespace",
 		"k8s_object_type",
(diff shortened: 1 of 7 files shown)


internal/utils/aggregator.go
```diff
@@ -40,6 +42,33 @@ func Aggregate_data(df dataframe.DataFrame) dataframe.DataFrame {
 
 	df = df.Mutate(s.Col("X0")).Rename("k8s_object_type", "X0")
 	df = df.Mutate(s.Col("X1")).Rename("k8s_object_name", "X1")
+
+	// filter out only valid workload type
+	df = df.Filter(
```
upadhyeammit, line 47: Is it possible to get this out as separate function? I understand its just a call to Filter however the `Aggregate_data` is already lengthy. With that can we have tests then? So looking at those one would understand what is invalid?

internal/utils/aggregator.go
```diff
@@ -40,6 +42,33 @@ func Aggregate_data(df dataframe.DataFrame) dataframe.DataFrame {
...
+				w.Replicaset.String(),
+				w.Replicationcontroller.String(),
+				w.Statefulset.String(),
+			}},
+	)
+
+	// Validation to check if metrics for cpuUsage, memoryUsage and memoryRSS are missing
+	df, no_of_dropped_records := filter_valid_csv_records(df)
+	if no_of_dropped_records != 0 {
+		invalidDataPoints.Add(float64(no_of_dropped_records))
```
upadhyeammit, line 64: If I understand then we are never going to return float from filter_valid_csv_records, is it prometheus which expects float value?

internal/utils/metrics.go
```diff
@@ -0,0 +1,13 @@
...
+
+import (
+	"github.com/prometheus/client_golang/prometheus"
+	"github.com/prometheus/client_golang/prometheus/promauto"
+)
+
+var (
+	invalidDataPoints = promauto.NewCounter(prometheus.CounterOpts{
+		Name: "rosocp_invalid_datapoints_total",
+		Help: "The total number of invalid datapoints(rows) found in CSVs recevied",
```
upadhyeammit, line 11: nit:
```suggestion
		Name: "rosocp_total_invalid_datapoints",
		Help: "The total number of invalid datapoints(rows) found in received CSVs",
```

internal/utils/aggregator.go
```diff
@@ -87,3 +116,25 @@ func Aggregate_data(df dataframe.DataFrame) dataframe.DataFrame {
 	df = dfGroups.Aggregation(columnsAggregationType, columnsToAggregate)
 	return df
 }
+
+func filter_valid_csv_records(main_df dataframe.DataFrame) (dataframe.DataFrame, int) {
```
upadhyeammit, line 120: If I understand if a value is missing then its 0 and so we can compare that to GreaterEq 0, I just like to confirm there is no missing or nil value?
At the same time can we have tests showcasing valid/invalid data?

internal/utils/aggregator.go
```diff
@@ -40,6 +42,33 @@ func Aggregate_data(df dataframe.DataFrame) dataframe.DataFrame {
...
+				w.Replicationcontroller.String(),
+				w.Statefulset.String(),
+			}},
+	)
+
+	// Validation to check if metrics for cpuUsage, memoryUsage and memoryRSS are missing
+	df, no_of_dropped_records := filter_valid_csv_records(df)
+	if no_of_dropped_records != 0 {
+		invalidDataPoints.Add(float64(no_of_dropped_records))
+		log.Infof("Invalid records in CSV - %v", no_of_dropped_records)
```
upadhyeammit, line 65: Does this need to be at Info level? and do you think we should also print more about owner_name and workload?

internal/utils/aggregator.go
```diff
@@ -40,6 +42,33 @@ func Aggregate_data(df dataframe.DataFrame) dataframe.DataFrame {
...
+		dataframe.F{
+			Colname:    "k8s_object_type",
+			Comparator: series.In,
+			Comparando: []string{
+				w.Daemonset.String(),
+				w.Deployment.String(),
+				w.Deploymentconfig.String(),
+				w.Replicaset.String(),
+				w.Replicationcontroller.String(),
+				w.Statefulset.String(),
```
saltgen, line 57: @patilsuraj767 Good one, no need to send every type to Kruize :+1:

internal/utils/aggregator.go
saltgen, line 65: Shall we remove this log statement? Details are getting captured via the Prometheus metric.

saltgen (changes requested):
@patilsuraj767 The Aggregate_data function seems quite long.
Suggestion: Could you split the function into 2, one could just filter the records and another aggregates?

internal/utils/aggregator.go
patilsuraj767: yes, prometheus `Add` method requires the float value.

internal/utils/metrics.go
patilsuraj767: According to prometheus metric naming convention - https://prometheus.io/docs/practices/naming/ suffix should describe the unit.

internal/utils/aggregator.go
patilsuraj767: Yes, good to have at info level so that using request_id we can check in kibana how many rows where dropped for particular request/CSV. Our logging system by default logs all the request related info - https://github.com/RedHatInsights/ros-ocp-backend/blob/main/internal/logging/logging.go#L66-L71

internal/utils/metrics.go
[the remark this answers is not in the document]
patilsuraj767: No, It will increase prometheus metrics cardinality.
We are logging the org_id, cluster_id, request_id etc to the cloudwatch.

internal/utils/aggregator.go
patilsuraj767: Please refer above comment - https://github.com/RedHatInsights/ros-ocp-backend/pull/144#discussion_r1381433018

saltgen: How would we cross reference `no_of_dropped_records` with the Prometheus metric, the metric count would gradually keep on increasing no?

saltgen: Please ignore

internal/utils/aggregator.go
patilsuraj767: We are filtering out/dropping the records where values are missing.

patilsuraj767: Added test cases

internal/utils/aggregator.go
patilsuraj767: Added test cases

upadhyeammit (approved):

saltgen (approved):
@patilsuraj767 LGTM :+1: Thank you for adding the tests :ok_hand:

PR #149, opened 2023-11-03
RHIROS-1411 Update recommendations job

saltgen:
### Job Steps

1. Checks if `/updateRecommendations` is up every 2 seconds with 10 attempts
2. Goes through every workload with missing recommendation and submits the respective `workload.ExperimentName` and `workload.MetricsUploadAt`
3. Saves recommendation and historical recommendation
4. Updates a Prometheus counter when invalid recommendations arrive from job execution

### DIFF
--- a/internal/jobs/update-recommendations.go
+++ b/internal/jobs/update-recommendations.go
@@ -1,0 +1,139 @@
+package jobs
+
+/*
+This job looks for missing recommendations, requests and saves them
+Originally intended for integration with Kruize v20
+*/
+
+import (
+	"encoding/json"
+	"fmt"
+	"net/url"
+	"reflect"
+	"time"
+
+	"github.com/prometheus/client_golang/prometheus"
+	"github.com/prometheus/client_golang/prometheus/promauto"
+
+	"github.com/redhatinsights/ros-ocp-backend/internal/config"
+	"github.com/redhatinsights/ros-ocp-backend/internal/logging"
+	"github.com/redhatinsights/ros-ocp-backend/internal/model"
+	"github.com/redhatinsights/ros-ocp-backend/internal/utils"
+	"github.com/redhatinsights/ros-ocp-backend/internal/utils/kruize"
+
+	database "github.com/redhatinsights/ros-ocp-backend/internal/db"
+)
+
+var log = logging.GetLogger()
+var cfg *config.Config = config.GetConfig()
+var db = database.GetDB()
+
+var (
+	invalidRecommendationUpdateRecommendationJob = promauto.NewCounter(prometheus.CounterOpts{
+		Name: "rosocp_invalid_recommendation_update_recommendation_job",
+		Help: "Invalid recommendations sum from Kruize gathered while executing update-recommendations job",
+	})
+)
+
+func checkURLStatus() bool {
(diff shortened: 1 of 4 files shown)


internal/jobs/update-recommendations.go
```diff
@@ -0,0 +1,145 @@
...
+	invalidRecommendationUpdateRecommendationJob = promauto.NewCounter(prometheus.CounterOpts{
+		Name: "rosocp_invalid_recommendation_update_recommendation_job",
+		Help: "Invalid recommendations sum from Kruize gathered while executing update-recommendations job",
+	})
+)
+
+func checkURLStatus() bool {
+
+	url := cfg.KruizeUrl + "/updateRecommendations"
+    resp, err := http.Get(url)
```
patilsuraj767, line 41: Have you tested this? I think kruize does not support `GET` method for` /updateRecommendations` API. So you will never get `200` response in return.

patilsuraj767 (changes requested):
Hello @saltgen code indentation in the PR is messed up can you please check.

internal/jobs/update-recommendations.go
```diff
@@ -0,0 +1,145 @@
...
+    // Process the results
+    for _, workload := range workloads {
+
+		recommendationSets, err := model.GetFirstRecommendationSetsByWorkloadID(workload.ID)
+		
+		if err != nil {
+			panic(err.Error())
+		}
+
+		if !reflect.ValueOf(recommendationSets).IsZero() {
```
patilsuraj767, line 84: Can you explain why there is a need to this check over here? and anyway we will be running this job after truncating the recommendationSets table that means this will every time evaluate to `False` and `else` block will be executed.

patilsuraj767 (changes requested):

upadhyeammit:
Instead of checking every 2 seconds for 10 attempts can we do: https://aws.amazon.com/blogs/architecture/exponential-backoff-and-jitter/

internal/jobs/update-recommendations.go
saltgen: I didn't want to repeat my testing cycles, additionally most of the code here is re-used. Hence the _Do Not Merge_ label.
Now that the new Kruize image is available, I'll get these minor aspects checked.

internal/jobs/update-recommendations.go
saltgen: - The exclamation symbol is more of a typo, I will revert that change.
- This check has been added to make the job idempotent, so that we can use this job in future if required.
- The recommendationSets table will be empty yes, but this job will look for recommendations for existing workloads, and then update those specifically

saltgen:
> Hello @saltgen code indentation in the PR is messed up can you please check.

I have no clue why the happens, locally formatting seems fine, will check.

internal/jobs/update-recommendations.go
saltgen: Change pushed

internal/jobs/update-recommendations.go
saltgen: Change pushed

saltgen:
Closing out this PR as stale

PR #153, opened 2023-11-16
Do not merge - test PR

chandrams:


pgarciaq:
## Implementation

Threaded the parent `ctx context.Context` through all recommendation functions in the ingest path.

### Changes

**`internal/services/report_processor.go`:**
- `runContainerRecommendations(ctx context.Context, kafkaMsg)` — accepts ctx instead of creating `context.Background()`
- `runNamespaceRecommendations(ctx context.Context, kafkaMsg)` — same
- `runStorageRecommendations(ctx context.Context, kafkaMsg)` — same
- `runSnapshotRecommendations(ctx context.Context, kafkaMsg)` — same
- `process*CSVNative` test helpers pass `context.Background()` explicitly (tests don't have a shutdown lifecycle)

**`internal/services/manifest_recommendations.go`:**
- `runClusterQuotaRecommendations(ctx context.Context, kafkaMsg)` — accepts ctx
- `runVMRecommendations(ctx context.Context, kafkaMsg)` — accepts ctx
- `runManifestRecommendations` now passes its ctx to each sub-function
- `runVMRecommendationsHook` type updated to include ctx

### Behavior change

When a SIGTERM is received during pod shutdown:
- **Before:** Recommendation functions continued DB transactions indefinitely using `context.Background()`, blocking shutdown and risking SIGKILL
- **After:** The Kafka handler's cancellable context propagates through, causing `pgx` queries to return `context.Canceled` promptly, allowing graceful drain within `terminationGracePeriodSeconds`

### Tests

All `internal/services/` tests pass. Test helpers continue using `context.Background()` since tests don't have a SIGTERM lifecycle.

Commit: d42b18db

PR #159, opened 2023-11-27
Upgrade kruize image to 0.0.19.5_rm

patilsuraj767:


patilsuraj767:
Merging the PR as there is no changes required from ROS side and pr_check is failing for some unknown reason which QE is aware of.

PR #162, opened 2023-12-02
Adjusting JAVA settings for 80% MaxRAMPercentage.

msvinaykumar:
Setting -XX:MaxRAMPercentage=80 in JAVA options adjusts the maximum RAM allocation to 80%. This means that when the Java Virtual Machine (JVM) runs, it will limit its memory usage to 80% of the available system RAM. This setting can be beneficial for resource management, preventing excessive memory usage and potential performance issues. It provides a balance between allowing Java applications to utilize a substantial amount of RAM while ensuring there is still sufficient memory for other system processes and applications. Adjusting this parameter can contribute to a more efficient and stable performance of Java applications.

### DIFF
--- a/kruize-clowdapp.yaml
+++ b/kruize-clowdapp.yaml
@@ -29,6 +29,8 @@ objects:
             cpu: ${KRUIZE_CPU_LIMIT}
             memory: ${KRUIZE_MEMORY_LIMIT}
         env:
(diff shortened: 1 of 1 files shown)


kgaikwad (approved):
Merging this MR as Kruize service need this config change.

PR #171, opened 2023-12-13
Removed commenting on the PR

chandrams:


kgaikwad (commented):
As you mentioned, now jenkins job run is happening against PR raised to update kruize image. Do you also want to revert hardcoded values for `new_tag` [here](https://github.com/RedHatInsights/ros-ocp-backend/blob/242dc08739fc575e85d240d03a6f1ae53bf91608/.github/workflows/test-ros-kruize-rm.yml#L50)?

chandrams:
@kgaikwad - I want to test the comment doesn't affect the triggering of the jenkins job, will remove the new_tag setting  in a subsequent PR

kgaikwad (approved):
Legit.

PR #175, opened 2024-01-02
Fixes #RHINENG-6254 - Added support for using default CA cert trust s…

patilsuraj767:
…tore

Platform change - App changes needed to prepare for AWS MSK (Kafka) migration

- [ ] Yes
- [x] No

- [ ] Bugfix
- [ ] New Feature
- [ ] Refactor
- [ ] Unittests Added
- [ ] DRY code
- [ ] Dependency Added
- [ ] DB Migration Added

## Additional :mega:


internal/kafka/consumer.go
```diff
@@ -25,14 +25,19 @@ func StartConsumer(kafka_topic string, handler func(msg *kafka.Message)) {
...
 			"enable.auto.commit":       cfg.KafkaAutoCommit,
 			"go.logs.channel.enable":   true,
 			"allow.auto.create.topics": true,
 		}
 
+		// As per librdkafka doc - https://github.com/confluentinc/librdkafka/blob/master/CONFIGURATION.md?plain=1#L73
+		// Default ca location is set to ca-certificates package. i.e. /etc/pki/ca-trust/extracted/pem/tls-ca-bundle.pem
+		if cfg.KafkaCA != "" {
+			configMap["ssl.ca.location"] = cfg.KafkaCA
+		}
```
upadhyeammit, line 39: I am bit confused! We are setting the KafkaCA using viper https://github.com/RedHatInsights/ros-ocp-backend/blob/main/internal/config/config.go#L92 ,
~~~
		if broker.Cacert != nil {
			caPath, err := c.KafkaCa(broker)
			if err != nil {
				panic("Kafka CA failed to write")
			}
			viper.Set("KafkaCA", caPath)
		}
~~~

So if its not nil we set it to caPath, I have below questions,
1. As per this PR do you expect caPath to be empty string and not the nil when actual path is not there? if yes then should we be handling it in config.go(initConfig()) itself?
2. If caPath can be nil, "" or actual path then also should this be handed in config.go(initConfig())?

patilsuraj767: Yes, if actual path is not present we get the empty string in `caPath`. 
I did not actually get what you mean by handling it in config.go(initConfig()). Does that mean we should shift creation of [kafka configMap](https://github.com/RedHatInsights/ros-ocp-backend/blob/main/internal/kafka/consumer.go#L21-L33) in config.go?

upadhyeammit: Nope, actually I mean/curious why we are again checking if cfg.KafkaCA is "" while creating configMap? and if we need to check it here then I think this can be checked in initConfig() itself? and I would expect just assignment without if condition in consumer.go,
~~~
configMap["ssl.ca.location"] = cfg.KafkaCA
~~~

upadhyeammit (approved):

patilsuraj767:
Thanks for the review @upadhyeammit. I am merging this PR.

PR #179, opened 2024-01-05
RHINENG-6226 Recommendation Poller service

saltgen:

Multiple recommendation requests due to chunked/multi-parted CSV reports are putting Kruize RDS under massive pressure.

Hence, a new service which concerns itself with looking for workloads which have sufficient data for yesterday and fetch recommendations for the same.

- [x] Yes (Pending)
- [ ] No

## Security Checklist :lock:

- [ ] Bugfix
- [x] New Feature
- [ ] Refactor
- [ ] Unittests Added
- [ ] DRY code
- [ ] Dependency Added
- [ ] DB Migration Added
- [x] EE Testing

### Steps to test locally

1. After staring the processor, execute `make run-recommendation-poller` in a new terminal window
2. Execute below commands in order,

- `make upload-msg-to-rosocp`
- `make upload-msg-to-rosocp-24Hrs`

Results should be similar to following;

Screenshot from 2024-01-05 21-43-56


internal/services/report_processor.go
```diff
@@ -174,71 +174,69 @@ func ProcessReport(msg *kafka.Message) {
...
-			recommendation, err := kruize.Update_recommendations(experiment_name, maxEndTime)
-			if err != nil {
-				end_interval := utils.ConvertDateToISO8601(maxEndTime.String())
-				if err.Error() == fmt.Sprintf("Recommendation for timestamp - \" %s \" does not exist", end_interval) {
-					log.Infof("Recommendation does not exist for timestamp - \" %s \"", end_interval)
+			maxEndtimeFromDB := recommendation_stored_in_db.MonitoringEndTime.UTC()
+			maxEndtimeFromReport := maxEndTime.UTC()
+			duration := maxEndtimeFromReport.Sub(maxEndtimeFromDB)
+
+			if reflect.ValueOf(recommendation_stored_in_db).IsZero() {
```
upadhyeammit, line 187: Does this mean when there are no recommendations for workload in db without checking if it has passed 24 hours we try to get recommendation for that? If yes then this should also be happening after 24 hours?

internal/services/recommendation_poller.go
```diff
@@ -0,0 +1,126 @@
...
+		return
+	}
+	if err := json.Unmarshal(msg.Value, &kafkaMsg); err != nil {
+		log.Errorf("unable to decode kafka message: %s", msg.Value)
+		return
+	}
+	if err := validate.Struct(kafkaMsg); err != nil {
+		log.Errorf("invalid kafka message: %s", err)
+		return
+	}
```
upadhyeammit, line 34: I can see this is similar in other processor as well, should this be moved to some common file?

internal/services/recommendation_poller.go
```diff
@@ -0,0 +1,126 @@
...
+	var isNewRecord bool = kafkaMsg.Metadata.New_record
+
+	if isNewRecord {
+		poll_for_recommendation = true
+	} else {
+		recommendation_stored_in_db, err := model.GetFirstRecommendationSetsByWorkloadID(workloadID)
+		if err != nil {
+			log.Errorf("Error while checking for recommendation_set record: %s", err)
+		}
+		duration := maxEndTimeFromReport.Sub(recommendation_stored_in_db.MonitoringEndTime.UTC())
```
upadhyeammit, line 54: Do we again need to depend on duration here? As we already have filtered it from report_processor?

clowdapp.yaml
```diff
@@ -151,7 +201,7 @@ objects:
     kafkaTopics:
       - topicName: hccm.ros.events
         partitions: 1
-      - topicName: rosocp.kruize.experiments
+      - topicName: rosocp.kruize.recommendations
         partitions: 1
```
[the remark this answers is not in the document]
patilsuraj767: partition settings here only take effect in EE. I can see in the platform-mq repo it is already set to 3 - [here](https://github.com/RedHatInsights/platform-mq/blob/master/topics/topics.json#L547).

internal/kafka/consumer.go
```diff
@@ -12,9 +12,17 @@ import (
 	"github.com/redhatinsights/ros-ocp-backend/internal/logging"
 )
 
-func StartConsumer(kafka_topic string, handler func(msg *kafka.Message)) {
+func StartConsumer(kafka_topic string, handler func(msg *kafka.Message, consumer_object *kafka.Consumer), auto_commit_option ...bool) {
```
patilsuraj767, line 15: Instead of changing function signature do you think it would be easy to have config variable which can be tweak by setting environment variable. Default value of this variable can be set to true and then on line 39 and 48 we can do something like `"enable.auto.commit": cfg.AUTOCOMMIT `

patilsuraj767 (changes requested):

internal/kafka/consumer.go
saltgen: We could go with config option, but updating the function signature would help us with reusing the consumer function programmatically.

In this specific case the consumer object can be passed onto the handler, so in future we may want this for some handler-specific kafka functionality for instance`consumer_object.CommitMsg`.

clowdapp.yaml
```diff
@@ -165,6 +215,9 @@ parameters:
 - description: Replica count for processor pod
   name: PROCESSOR_REPLICA_COUNT
   value: "1"
+- description: Replica count for recommendation-poller pod
+  name: POLLER_REPLICA_COUNT
+  value: "1"
```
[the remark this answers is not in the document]
saltgen: We generally bump replica count in the sre repo, this is for EE

internal/config/config.go
```diff
@@ -12,17 +12,18 @@ import (
...
 type Config struct {
-	//Application config
-	ServiceName string `mapstructure:"SERVICE_NAME"`
-	LogFormater string `mapstructure:"LogFormater"`
-	LogLevel    string `mapstructure:"LOG_LEVEL"`
+	// Application config
+	ServiceName              string `mapstructure:"SERVICE_NAME"`
+	LogFormater              string `mapstructure:"LogFormater"`
+	LogLevel                 string `mapstructure:"LOG_LEVEL"`
+	RecommendationFetchDelay int    `mapstructure:"RECOMMENDATION_FETCH_DELAY"`
```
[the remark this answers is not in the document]
saltgen: Config name updated

internal/kafka/consumer.go
```diff
@@ -12,9 +12,17 @@ import (
...
 	log := logging.GetLogger()
 	cfg := config.GetConfig()
+
+	var auto_commit bool
+	if len(auto_commit_option) > 0 {
+		auto_commit = auto_commit_option[0]
+	} else {
+		auto_commit = cfg.KafkaAutoCommit
+	}
```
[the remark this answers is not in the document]
saltgen: `auto_commit` is being used as an option for respective consumer, more details [here](https://github.com/RedHatInsights/ros-ocp-backend/pull/179#discussion_r1445961678)

internal/services/report_processor.go
```diff
@@ -174,71 +174,69 @@ func ProcessReport(msg *kafka.Message) {
...
-				}
 			}
 
-			recommendation, err := kruize.Update_recommendations(experiment_name, maxEndTime)
-			if err != nil {
-				end_interval := utils.ConvertDateToISO8601(maxEndTime.String())
-				if err.Error() == fmt.Sprintf("Recommendation for timestamp - \" %s \" does not exist", end_interval) {
-					log.Infof("Recommendation does not exist for timestamp - \" %s \"", end_interval)
+			maxEndtimeFromDB := recommendation_stored_in_db.MonitoringEndTime.UTC()
+			maxEndtimeFromReport := maxEndTime.UTC()
```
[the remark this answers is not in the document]
saltgen: Yes it should, but I'd rather refactor processor related changes later.
As an excuse, using maxEndtimeFromReport for explicit UTC conversion atm :grin:

internal/services/report_processor.go
```diff
@@ -174,71 +174,69 @@ func ProcessReport(msg *kafka.Message) {
...
+			duration := maxEndtimeFromReport.Sub(maxEndtimeFromDB)
```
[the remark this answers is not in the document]
saltgen: Yes, committing kafka msg in such scenario

internal/services/recommendation_poller.go
saltgen: Yes, it should be moved. As discussed, we'll will look into the import cycle error w.r.t utils later

internal/services/recommendation_poller.go
saltgen: We needed to, for graceful msg handling, but current changes should resolve this query better

saltgen:
Please excuse the lint related commits, I am facing some local issues while linting.

internal/services/recommendation_poller.go
```diff
@@ -0,0 +1,141 @@
...
+	}
+	if err := json.Unmarshal(msg.Value, &kafkaMsg); err != nil {
+		log.Errorf("unable to decode kafka message: %s", msg.Value)
+		commitKafkaMsg(msg, consumer_object)
+		return
+	}
+	if err := validate.Struct(kafkaMsg); err != nil {
+		log.Errorf("invalid kafka message: %s", err)
+		commitKafkaMsg(msg, consumer_object)
+		return
```
saltgen, line 107: @patilsuraj767 Would these `return`s stop the consumer instance? I have replicated these from the processor.
Looking for a mechanism to stop code flow post `commitKafkaMsg` in this handler.

internal/services/recommendation_poller.go
```diff
@@ -15,112 +16,126 @@ import (
...
+				// Create RecommendationSet entry into the table.
+				recommendationSet := model.RecommendationSet{
+					WorkloadID:          kafkaMsg.Metadata.Workload_id,
+					ContainerName:       container.Container_name,
+					MonitoringStartTime: v.Duration_based.Short_term.Monitoring_start_time,
+					MonitoringEndTime:   v.Duration_based.Short_term.Monitoring_end_time,
+					Recommendations:     marshalData,
+				}
+				if err := recommendationSet.CreateRecommendationSet(); err != nil {
+					log.Errorf("unable to save a record into recommendation set: %v. Error: %v", recommendationSet, err)
```
upadhyeammit, line 60: I think if we are unable to save the recommendations then also we should not commit the message?

internal/services/recommendation_poller.go
```diff
@@ -15,112 +16,126 @@ import (
...
+					OrgId:               kafkaMsg.Metadata.Org_id,
+					WorkloadID:          kafkaMsg.Metadata.Workload_id,
+					ContainerName:       container.Container_name,
+					MonitoringStartTime: v.Duration_based.Short_term.Monitoring_start_time,
+					MonitoringEndTime:   v.Duration_based.Short_term.Monitoring_end_time,
+					Recommendations:     marshalData,
+				}
+				if err := historicalRecommendationSet.CreateHistoricalRecommendationSet(); err != nil {
+					recommendationJSON, _ := json.Marshal(recommendation)
+					log.Errorf("unable to get or add record to historical recommendation set table: %s. Error: %v", string(recommendationJSON), err)
```
upadhyeammit, line 76: Same here as well? I think if we are unable to save the recommendations then also we should not commit the message?

upadhyeammit (commented):
Just added two comments, I will again check this tomorrow with fresh mind!

internal/services/recommendation_poller.go
saltgen: I do agree, relocated poll_cycle_complete, feel free to take a look

internal/services/recommendation_poller.go
saltgen: Ditto

internal/services/recommendation_poller.go
upadhyeammit: @saltgen if I see the whole code block now below is my understanding,
1. If CreateRecommendationSet() and CreateHistoricalRecommendationSet() are successful 
    - We commit the message.
2. If CreateRecommendationSet() and CreateHistoricalRecommendationSet() are unsuccessful 
    - We dont commit the message.
3. If CreateRecommendationSet() is successful and CreateHistoricalRecommendationSet() is unsuccessful
    - We don't commit the message
4. If CreateRecommendationSet() is unsuccessful and CreateHistoricalRecommendationSet() is successful
    - We don't commit the message

So in case 3 and 4 we are able to save the data in one of the table but not both. In this situation if we don't commit and when next time we would again handle same message,
A. If CreateRecommendationSet() failed in first iteration this means would be able to save it because the end time is still old in database.
B. If CreateHistoricalRecommendationSet() failed in first iteration this means the end timestamp in db is now same as kafka message and this mean we won't be able to do CreateHistoricalRecommendationSet().

Is it okay if we miss to save in HistoricalRecommendation table in this situation?

upadhyeammit: If we wish to either success or fail for both then I think we can do https://gorm.io/docs/transactions.html

internal/services/recommendation_poller.go
```diff
@@ -0,0 +1,176 @@
...
+				}
+				// Create RecommendationSet entry into the table.
+				recommendationSet := model.RecommendationSet{
+					WorkloadID:          kafkaMsg.Metadata.Workload_id,
+					ContainerName:       container.Container_name,
+					MonitoringStartTime: v.Duration_based.Short_term.Monitoring_start_time,
+					MonitoringEndTime:   v.Duration_based.Short_term.Monitoring_end_time,
+					Recommendations:     marshalData,
+				}
+				recommendationSetList = append(recommendationSetList, recommendationSet)
```
saltgen, line 96: Added change!
Per container there will be one recomm. so best to end transaction/poll cycle after each recomm. per container has been saved.

internal/services/recommendation_poller.go
saltgen: DB transaction added

internal/services/recommendation_poller.go
saltgen: Ditto

internal/kafka/consumer.go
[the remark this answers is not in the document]
saltgen: https://github.com/RedHatInsights/ros-ocp-backend/pull/179#discussion_r1450209665

internal/services/recommendation_poller.go
```diff
@@ -0,0 +1,176 @@
...
+		return
+	}
+	log = logging.Set_request_details_recommendations(kafkaMsg)
+
+	maxEndTimeFromReport := kafkaMsg.Metadata.Max_endtime_report
+	workloadID := kafkaMsg.Metadata.Workload_id
+
+	recommendation_stored_in_db, err := model.GetFirstRecommendationSetsByWorkloadID(workloadID)
+	if err != nil {
+		log.Errorf("Error while checking for recommendation_set record: %s", err)
```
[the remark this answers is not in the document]
saltgen: Thanks! change pushed

patilsuraj767 (approved):
Thanks @saltgen, I have tested changes locally looks good to me 👍

upadhyeammit (approved):

PR #180, opened 2024-01-08
Fixes #RHINENG-2879 - Pin Golang version for go-toolset

upadhyeammit:

The latest tag of go-toolset keeps on updating as per the 1.y versions of Go. Considering the go packages being used in repository are tested for Golang 1.19.z this change specifically pins it in Dockerfile to be surefooted for production releases.

- [ ] Yes
- [ ] No

## Security Checklist :lock:

- [ ] Bugfix
- [ ] New Feature
- [ ] Refactor
- [ ] Unittests Added
- [ ] DRY code
- [ ] Dependency Added
- [ ] DB Migration Added

## Additional :mega:

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,13 +1,15 @@
-FROM registry.redhat.io/ubi8/go-toolset:latest as builder
+FROM registry.redhat.io/ubi8/go-toolset:1.20 as builder
 WORKDIR /go/src/app
 COPY . .
 USER 0
 RUN go get -d ./... && \
-    go build -o rosocp rosocp.go
+    go build -o rosocp rosocp.go && \
+    echo "$(go version)" > go_version_details
 
 FROM registry.redhat.io/ubi8/ubi-minimal:latest
 WORKDIR /
 COPY --from=builder /go/src/app/rosocp ./rosocp
+COPY --from=builder /go/src/app/go_version_details ./go_version_details
 COPY migrations ./migrations
 COPY openapi.json ./openapi.json
 COPY resource_optimization_openshift.json ./resource_optimization_openshift.json
--- a/go.mod
+++ b/go.mod
@@ -1,6 +1,6 @@
 module github.com/redhatinsights/ros-ocp-backend
 
-go 1.19
+go 1.20
 
 require (
 	github.com/aws/aws-sdk-go v1.38.51


upadhyeammit:
Its not possible to set the labels using bash command! Considering this I have created new file which has the go version. Pinning the go version in image solves half of the problem as under 1.19 label there could be older or newer patch versions(1.19.13/1.19.12 etc)

Dockerfile
```diff
@@ -1,13 +1,15 @@
-FROM registry.redhat.io/ubi8/go-toolset:latest as builder
+FROM registry.redhat.io/ubi8/go-toolset:1.19 as builder
 WORKDIR /go/src/app
 COPY . .
 USER 0
 RUN go get -d ./... && \
-    go build -o rosocp rosocp.go
+    go build -o rosocp rosocp.go && \
+    echo "$(go version)" > go_version_details
```
saltgen, line 7: I don't think we'll need to do this if specific tag is mentioned for go-toolset

saltgen (changes requested):

Dockerfile
saltgen: We could pin the patch/subversion tag as well

Dockerfile
```diff
@@ -1,13 +1,15 @@
-FROM registry.redhat.io/ubi8/go-toolset:latest as builder
+FROM registry.redhat.io/ubi8/go-toolset:1.19 as builder
```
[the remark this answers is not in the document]
upadhyeammit: I wish to upgrade the go mod file version to 1.20 as well if we wish to go for 1.20. So should I do that?

Dockerfile
upadhyeammit: Yes, we can, however that can be overkill considering we would end up not getting the fixes by default in patched versions.

saltgen:
@upadhyeammit Let's get these changes tested on EE, happy to help you with the minio hack + any other EE setup

patilsuraj767:
> @upadhyeammit Let's get these changes tested on EE, happy to help you with the minio hack + any other EE related setup

I can see pr_checks job has passed, So new image should had been already tested in EE as a part of pr_checks.

patilsuraj767 (approved):
Thanks @upadhyeammit LGTM 👍

saltgen (approved):
Thanks @upadhyeammit for the changes :100:

PR #184, opened 2024-02-06
Fixes #RHINENG-7893 - accept and serve boxplots

upadhyeammit:

The Kruize added feature to include the boxplot data in response. The ROS should accept the new boxplot data, save it and serve it finally to be consumed by UI/API.

- [ ] Yes
- [x] No

## Security Checklist :lock:

- [x] Does this change depend on specific version of Kruize
  - Kruize version is undecided as of now
- [ ] Bugfix
- [x] New Feature
- [ ] Refactor
- [ ] Unittests Added
- [ ] DRY code
- [ ] Dependency Added
- [ ] DB Migration Added

## Additional :mega:

### DIFF
--- a/internal/api/utils.go
+++ b/internal/api/utils.go
@@ -194,10 +194,11 @@ func get_user_permissions(c echo.Context) map[string][]string {
 	return user_permissions
 }
 
-func transformComponentUnits(recommendationJSON map[string]interface{}) map[string]interface{} {
+func truncateComponentUnits(recommendationJSON map[string]interface{}) map[string]interface{} {
 	/*
 		Truncates CPU units(cores) to three decimal places
 		Truncates Memory units(Mi) to two decimal places
+		Truncates duration_in_hours to one decimal places
 	*/
 
 	hasMoreThanThreeDecimals := func(value float64) bool {
@@ -232,6 +233,14 @@ func transformComponentUnits(recommendationJSON map[string]interface{}) map[stri
 		return ok
 	}
 
+	truncateDurationInHours := func(intervalData map[string]interface{}) bool {
+		durationInHours, ok := intervalData["duration_in_hours"].(float64)
+		if ok {
+			intervalData["duration_in_hours"] = math.Trunc(durationInHours*10) / 10
+		}
+		return ok
+	}
+
 	// Current section of recommendation
 	current_config, ok := recommendationJSON["current"].(map[string]interface{})
 	if !ok {
@@ -246,7 +255,7 @@ func transformComponentUnits(recommendationJSON map[string]interface{}) map[stri
 			if ok {
 				err := truncateMemory(memInMi)
 				if !err {
-					fmt.Printf("error truncating the memory in %s: %v\n", sectionObject, err)
+					log.Errorf("error truncating the memory in %s\n", sectionObject)
 					continue
 				}
 			}
@@ -255,7 +264,7 @@ func transformComponentUnits(recommendationJSON map[string]interface{}) map[stri
 			if ok {
 				err := truncateCPU(cpu)
 				if !err {
-					fmt.Printf("error truncating cpu in %s: %v\n", sectionObject, err)
+					log.Errorf("error truncating cpu in %s\n", sectionObject)
 					continue
 				}
 			}
@@ -292,6 +301,11 @@ func transformComponentUnits(recommendationJSON map[string]interface{}) map[stri
 			delete(intervalData, "monitoring_start_time")
 		}
 
+		err := truncateDurationInHours(intervalData)
+		if !err {
+			log.Errorf("error truncating duration_in_hours in term %s\n", period)
+		}
+
 		if intervalData["recommendation_engines"] != nil {
 
 			for _, recommendationType := range []string{"cost", "performance"} {
@@ -314,7 +328,7 @@ func transformComponentUnits(recommendationJSON map[string]interface{}) map[stri
 							if ok {
 								err := truncateMemory(memInMi)
 								if !err {
-									fmt.Printf("error truncating the memory in %s: %v\n", period, err)
+									log.Errorf("error truncating the memory in %s\n", period)
 									continue
 								}
 							}
@@ -323,7 +337,7 @@ func transformComponentUnits(recommendationJSON map[string]interface{}) map[stri
 							if ok {
 								err := truncateCPU(cpu)
 								if !err {
-									fmt.Printf("error truncating cpu in %s: %v\n", period, err)
+									log.Errorf("error truncating cpu in %s\n", period)
 									continue
 								}
 							}
@@ -397,7 +411,7 @@ func UpdateRecommendationJSON(recommendationID string, clusterUUID string, jsonD
 		return nil
 	}
 
-	data = transformComponentUnits(data)
+	data = truncateComponentUnits(data)
 	data = filterNotifications(recommendationID, clusterUUID, data)
 	return data
 }
--- a/internal/api/utils.go
+++ b/internal/api/utils.go
@@ -199,7 +199,7 @@ func truncateComponentUnits(recommendationJSON map[string]interface{}) map[strin
 		Truncates CPU units(cores) to three decimal places
 		Truncates Memory units(Mi) to two decimal places
 		Hack: Truncates duration_in_hours to one decimal places
-		To Do: Once Kruize returns identical values for duration_in_hours
+		TODO: Once Kruize returns identical values for duration_in_hours
 		the ros-ocp should stop truncating the duration_in_hours
 	*/
 
--- a/internal/api/utils.go
+++ b/internal/api/utils.go
@@ -198,7 +198,9 @@ func truncateComponentUnits(recommendationJSON map[string]interface{}) map[strin
 	/*
 		Truncates CPU units(cores) to three decimal places
 		Truncates Memory units(Mi) to two decimal places
-		Truncates duration_in_hours to one decimal places
+		Hack: Truncates duration_in_hours to one decimal places
+		To Do: Once Kruize returns identical values for duration_in_hours
+		the ros-ocp should stop truncating the duration_in_hours
 	*/
 
 	hasMoreThanThreeDecimals := func(value float64) bool {
--- a/internal/api/utils.go
+++ b/internal/api/utils.go
@@ -196,7 +196,8 @@ func get_user_permissions(c echo.Context) map[string][]string {
 
 func transformComponentUnits(recommendationJSON map[string]interface{}) map[string]interface{} {
 	/*
-		Truncates CPU units to three decimal places
+		Truncates CPU units(cores) to three decimal places
+		Truncates Memory units(Mi) to two decimal places
 	*/
 
 	hasMoreThanThreeDecimals := func(value float64) bool {
@@ -223,6 +224,14 @@ func transformComponentUnits(recommendationJSON map[string]interface{}) map[stri
 		return ok
 	}
 
+	truncateMemory := func(mem map[string]interface{}) bool {
+		memory, ok := mem["amount"].(float64)
+		if ok {
+			mem["amount"] = math.Trunc(memory*100) / 100
+		}
+		return ok
+	}
+
 	// Current section of recommendation
 	current_config, ok := recommendationJSON["current"].(map[string]interface{})
 	if !ok {
@@ -233,11 +242,20 @@ func transformComponentUnits(recommendationJSON map[string]interface{}) map[stri
 
 		sectionObject, ok := current_config[section].(map[string]interface{})
 		if ok {
+			memInMi, ok := sectionObject["memory"].(map[string]interface{})
+			if ok {
+				err := truncateMemory(memInMi)
+				if !err {
+					fmt.Printf("error truncating the memory in %s: %v\n", sectionObject, err)
+					continue
+				}
+			}
+
 			cpu, ok := sectionObject["cpu"].(map[string]interface{})
 			if ok {
 				err := truncateCPU(cpu)
 				if !err {
-					fmt.Printf("error converting cpu in %s: %v\n", sectionObject, err)
+					fmt.Printf("error truncating cpu in %s: %v\n", sectionObject, err)
 					continue
 				}
 			}
@@ -292,6 +310,15 @@ func transformComponentUnits(recommendationJSON map[string]interface{}) map[stri
 
 						sectionObject, ok := recommendationSection[section].(map[string]interface{})
 						if ok {
+							memInMi, ok := sectionObject["memory"].(map[string]interface{})
+							if ok {
+								err := truncateMemory(memInMi)
+								if !err {
+									fmt.Printf("error truncating the memory in %s: %v\n", period, err)
+									continue
+								}
+							}
+
 							cpu, ok := sectionObject["cpu"].(map[string]interface{})
 							if ok {
 								err := truncateCPU(cpu)
--- a/internal/api/utils.go
+++ b/internal/api/utils.go
@@ -196,26 +196,8 @@ func get_user_permissions(c echo.Context) map[string][]string {
 
 func transformComponentUnits(recommendationJSON map[string]interface{}) map[string]interface{} {
 	/*
-		Converts units for Memory and CPU
-		bytes -> MiB -> GiB
-		cores -> millicores
+		Truncates CPU units to three decimal places
 	*/
-
-	convertMemory := func(memory map[string]interface{}) error {
-		amount, ok := memory["amount"].(float64)
-		if ok {
-			memoryInMiB := amount / 1024 / 1024
-			if math.Abs(memoryInMiB) >= 1024 {
-				memoryInGiB := memoryInMiB / 1024
-				memory["amount"] = math.Trunc(memoryInGiB*100) / 100
-				memory["format"] = "Gi"
-			} else {
-				memory["amount"] = math.Trunc(memoryInMiB*100) / 100
-				memory["format"] = "Mi"
-			}
-		}
-		return nil
-	}
 
 	hasMoreThanThreeDecimals := func(value float64) bool {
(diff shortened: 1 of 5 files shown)


upadhyeammit:
FYI: I have kept the docker-compose.yml change in commit as of now, so its easy to test locally. Before merging the pr we need to remove the change of docker-compose.yml

openapi.json
```diff
@@ -430,6 +430,9 @@
                 }
               }
             }
+          },
+          "plots":{
+            "$ref": "#/components/schemas/LongTermPlots"
```
patilsuraj767, line 435:
```suggestion
            "$ref": "#/components/schemas/TermPlots"
```
Instead of defining different term plots like `LongTermPlots`, `MediumTermPlots` and `ShortTermPlots`. Can we DRY principle only have one like `TermPlots`.

openapi.json
```diff
@@ -466,6 +469,9 @@
                 }
               }
             }
+          },
+          "plots":{
+            "$ref":"#/components/schemas/MediumTermPlots"
```
patilsuraj767, line 474:
```suggestion
            "$ref":"#/components/schemas/TermPlots"
```

openapi.json
```diff
@@ -530,6 +536,9 @@
                 "$ref": "#/components/schemas/PerformanceRecommendation"
               }
             }
+          },
+          "plots":{
+            "$ref": "#/components/schemas/ShortTermPlots"
```
patilsuraj767, line 541:
```suggestion
            "$ref": "#/components/schemas/TermPlots"
```

openapi.json
```diff
@@ -1554,6 +1563,401 @@
             "example": "deploymentconfig"
           }
         }
+      },
+      "ShortTermPlots":{
```
patilsuraj767, line 1567:
```suggestion
      "TermPlots":{
```

openapi.json
```diff
@@ -1554,6 +1563,401 @@
...
+        "properties":{
+          "datapoints":{
+            "type": "integer",
+            "example": 15
+          },
+          "plots_data":{
+            "$ref": "#/components/schemas/LongTermPlotsData"
+          }
+        }
+      },
```
patilsuraj767, line 1602:
```suggestion
     
```

openapi.json
```diff
@@ -1554,6 +1563,401 @@
...
+      "ShortTermPlots":{
+        "type": "object",
+        "properties":{
+          "datapoints":{
+            "type": "integer",
+            "example": 4
+          },
+        "plots_data":{
+          "$ref": "#/components/schemas/ShortTermPlotsData"
```
patilsuraj767, line 1575:
```suggestion
          "$ref": "#/components/schemas/TermPlotsData"
```
Same

openapi.json
```diff
@@ -1554,6 +1563,401 @@
...
+          "datapoints":{
+            "type": "integer",
+            "example": 15
+          },
+          "plots_data":{
+            "$ref": "#/components/schemas/LongTermPlotsData"
+          }
+        }
+      },
+      "ShortTermPlotsData": {
```
patilsuraj767, line 1603:
```suggestion
      "TermPlotsData": {
```

openapi.json
```diff
@@ -1554,6 +1563,401 @@
...
+              "cpuUsage":{
+                "$ref": "#/components/schemas/cpuUsage"
+              },
+              "memoryUsage":{
+                "$ref":"#/components/schemas/memoryUsage"
+              }
+            }
+          }
+        }
+      },
```
patilsuraj767, line 1903:
```suggestion
      
```

patilsuraj767 (dismissed):

openapi.json
upadhyeammit: I was willing to do the same however for each term number of datapoints differ as example for short term its 4, medium term its 7 and long term its 15. At the same time timestamps differ, as example for short term its every 6 hours, for medium and long term its every day. So to match API response I had to keep these separate.

patilsuraj767: Ok got it. In that case we can keep components `LongTermPlots`, `MediumTermPlots` and `ShortTermPlots`.
Can we change `ShortTermPlotsData`, `MediumTermPlotsData`, `LongTermPlotsData` to single component `TermPlotsData`?

kgaikwad:
This PR will also need a change https://github.com/RedHatInsights/ros-ocp-backend/pull/201 to merge

openapi.json
upadhyeammit: I think no, because every `ShortTermPlotsData`, `MediumTermPlotsData` and `LongTermPlotsData` have different number of properties with varying timestamps?

patilsuraj767: I see only `timestamp` is different but other two `cpuUsage` and `memoryUsage` are the same. I think we can keep the same timestamp for all the examples.

upadhyeammit: Yeah we can, however then it does not depict exact response?

saltgen (changes requested):

openapi.json
upadhyeammit: As per discussion with Sagnik its not needed match exact api response in openapi. With that I reduced the file, and it has far less duplicate details now.

saltgen (changes requested):

internal/utils/convert_memory/convert_memory.go
saltgen, line 1: Adding ref to https://github.com/RedHatInsights/ros-ocp-backend/pull/184#discussion_r1572188446

internal/types/kruizePayload/common.go
```diff
@@ -256,3 +281,23 @@ func make_container_data(c map[string]interface{}) container {
...
+func MemBytesToMi(mem float64) float64 {
+	memoryInMi := mem / 1024 / 1024
+	truncatedMem := math.Trunc(memoryInMi*100) / 100
+	return truncatedMem
+}
+
+func isMemoryField(metricName string) bool {
+	// TO DO: In go 1.21 use slices package instead of map
+	memoryFields := map[string]string{"memoryRSS": "", "memoryUsage": "", "memoryLimit": "", "memoryRequest": ""}
+	_, keyExists := memoryFields["metricName"]
```
saltgen, line 294: I see some bad memory values on API response, maybe this?
Seems like `metricName string` is already a string.

kgaikwad:
@upadhyeammit  - https://github.com/RedHatInsights/ros-ocp-backend/pull/201 is merged, you can rebase this branch to verify your change with updated Kruize image. Thanks!

upadhyeammit:
> @upadhyeammit - #201 is merged, you can rebase this branch to verify your change with updated Kruize image. Thanks!

Done, its working!

internal/types/kruizePayload/common.go
```diff
@@ -282,22 +291,15 @@ func make_container_data(c map[string]interface{}) container {
...
-	memoryFields := map[string]string{"memoryRSS": "", "memoryUsage": "", "memoryLimit": "", "memoryRequest": ""}
-	_, keyExists := memoryFields[metricName]
-	return keyExists
-}
-
-func convertMemoryFields(metricName string, field string, c map[string]interface{}) {
-	if isMemoryField(metricName) {
-		converted_mem := MemBytesToMi(c[field].(float64))
-		c[field] = converted_mem
+func convertMemoryFields(field string, c map[string]interface{}) {
```
upadhyeammit, line 294: I have removed the Trunc usage from here as even if we Trunc the memory to 3 decimal points kruize returns it with more decimals.
At the same time if we don't update the c[field] in this function then we would end up repeating the below logic multiple times,
~~~
someField, ok := c[field].(float64)
if ok {
do the convert
}
else{
Warn
}
~~~

This is because go does not allow us to pass `c[field]` directly, it needs to be cast first and that would mean we cast it everytime before we pass it to func. So I kept the logic as is, and this also resembles to what we do at https://github.com/RedHatInsights/ros-ocp-backend/blob/main/internal/api/utils.go#L231 only difference is here its separate func.

openapi.json
```diff
@@ -1502,6 +1513,126 @@
...
+            "$ref": "#/components/schemas/cpuUsageFloatComponent"
+          }
+        }
+      },
+      "memoryUsage":{
+        "type": "object",
+        "properties":{
+          "format":{
+            "type": "string",
+            "example": "Mi"
```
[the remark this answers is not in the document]
saltgen: @upadhyeammit Additionally `duration_in_hours` using sample csv for long_term and short_term, seem to have too much precision

internal/types/kruizePayload/common.go
```diff
@@ -256,3 +290,16 @@ func make_container_data(c map[string]interface{}) container {
 
 	return container_data
 }
+
+func convertMemoryFields(field string, c map[string]interface{}) {
```
saltgen, line 294:
```suggestion
func convertMemoryField(field string, c map[string]interface{}) {
```

internal/api/utils.go
```diff
@@ -324,19 +310,20 @@ func transformComponentUnits(recommendationJSON map[string]interface{}) map[stri
...
 							}
+
 							cpu, ok := sectionObject["cpu"].(map[string]interface{})
 							if ok {
-								err := convertCPU(cpu)
-								if err != nil {
-									fmt.Printf("error converting cpu in %s: %v\n", period, err)
+								err := truncateCPU(cpu)
+								if !err {
+									fmt.Printf("error truncating cpu in %s: %v\n", period, err)
```
saltgen, line 326: If possible, please help with convert this `fmt.Printf` statements to `log.Error` 🙂

saltgen (changes requested):

internal/types/kruizePayload/common.go
saltgen: Instead of setting c[field] = memoryInMi explicitly, we can capture the value of
convertMemoryFields function if it returns the same.

convertedValue := convertMemoryFields(max_field, c)

// Assign the max value returned
max = AssertAndConvertToString(convertedValue)

openapi.json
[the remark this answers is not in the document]
upadhyeammit: I have confirmed with Vinay and as per that format will stay same with or without data, that means we expect `Mi` as format from kruize even when everything else is just zero.

Yeah, just spotted the precision of `duration_in_hours, let me check and limit it to double digit only!

internal/types/kruizePayload/common.go
upadhyeammit: Okay, I tried to give it a shot however then `AssertAndConvertToString` won't work for cpu and mem, it won't be generalized func? Or I will need to change float64->interface and then AssertAndConvertToString can handle it?

openapi.json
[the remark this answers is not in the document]
upadhyeammit: I truncated duration_in_hours to 1 decimal as for hours it suits better!

internal/types/kruizePayload/common.go
saltgen: At the moment `AssertAndConvertToString` seems to handle `float`, `int` and `string`, and looks like a general usage function

internal/api/utils.go
```diff
@@ -194,29 +194,13 @@ func get_user_permissions(c echo.Context) map[string][]string {
...
 
-func transformComponentUnits(recommendationJSON map[string]interface{}) map[string]interface{} {
+func truncateComponentUnits(recommendationJSON map[string]interface{}) map[string]interface{} {
 	/*
-		Converts units for Memory and CPU
-		bytes -> MiB -> GiB
-		cores -> millicores
+		Truncates CPU units(cores) to three decimal places
+		Truncates Memory units(Mi) to two decimal places
+		Truncates duration_in_hours to one decimal places
```
saltgen, line 201:
```suggestion
		HACK: Truncates duration_in_hours to one decimal places
```

saltgen (changes requested):

internal/types/kruizePayload/common.go
```diff
@@ -72,6 +74,26 @@ type RecommendationTerm struct {
 		Cost        RecommendationEngineObject `json:"cost,omitempty"`
 		Performance RecommendationEngineObject `json:"performance,omitempty"`
 	} `json:"recommendation_engines,omitempty"`
+	Plots Plot `json:"plots,omitempty"`
+}
+
+type Plot struct {
+	DataPoints int                  `json:"datapoints,omitempty"`
+	PlotsData  map[string]PlotsData `json:"plots_data,omitempty"`
```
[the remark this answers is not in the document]
saltgen: Looks like the issue is not resolved yet, request you to compare locally.

```
 "long_term": {
                        "duration_in_hours": 0.9,
                        "plots": {
                            "datapoints": 15,
                            "plots_data": {
                                "2023-05-21T00:45:00.000Z": {
                                    "cpuUsage": {
                                        "format": "",
                                        "max": 0,
                                        "median": 0,
                                        "min": 0,
                                        "q1": 0,
                                        "q3": 0
                                    },
```

openapi.json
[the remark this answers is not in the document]
saltgen: The field shows incorrect values now for some reason 🤔 

```
 "long_term": {
                        "duration_in_hours": 0.9,
                        "plots": {
                            "datapoints": 15,
                            "plots_data": {

              
```

internal/api/utils.go
```diff
@@ -232,19 +218,29 @@ func transformComponentUnits(recommendationJSON map[string]interface{}) map[stri
...
+		}
+		return ok
+	}
+
+	truncateDurationInHours := func(intervalData map[string]interface{}) bool {
+		durationInHours, ok := intervalData["duration_in_hours"].(float64)
+		if ok {
+			intervalData["duration_in_hours"] = math.Trunc(durationInHours*10) / 10
+		}
+		return ok
```
saltgen, line 243: Noticed values as `int` for short term and `float` for other using sample csv, same applies for values in openapi spec

internal/api/utils.go
```diff
@@ -194,29 +194,15 @@ func get_user_permissions(c echo.Context) map[string][]string {
...
-func transformComponentUnits(recommendationJSON map[string]interface{}) map[string]interface{} {
+func truncateComponentUnits(recommendationJSON map[string]interface{}) map[string]interface{} {
 	/*
-		Converts units for Memory and CPU
-		bytes -> MiB -> GiB
-		cores -> millicores
+		Truncates CPU units(cores) to three decimal places
+		Truncates Memory units(Mi) to two decimal places
+		Hack: Truncates duration_in_hours to one decimal places
+		To Do: Once Kruize returns identical values for duration_in_hours
```
saltgen, line 202:
```suggestion
		TODO: Once Kruize returns identical values for duration_in_hours
```
Nitpick: Helpful to discover Todos across codebase by editor/IDE

saltgen (changes requested):

openapi.json
```diff
@@ -1502,6 +1513,126 @@
...
+          },
+        "plots_data":{
+          "$ref": "#/components/schemas/PlotDetails"
+          }
+        }
+      },
+      "PlotDetails": {
+        "type": "object",
+        "properties":{
+          "2023-04-01T06:00:00Z":{
```
saltgen, line 1532: Let's add an empty datapoint sample as well.
```
"plots_data": {
                                "2023-05-28T00:45:00.000Z": {},
                                "2023-05-29T00:45:00.000Z": {},
```

internal/api/utils.go
upadhyeammit: Updated the Openapi.json

openapi.json
```diff
@@ -1502,6 +1519,116 @@
...
+            "properties":{
+              "cpuUsage":{
+                "$ref": "#/components/schemas/cpuUsage"
+              },
+              "memoryUsage":{
+                "$ref":"#/components/schemas/memoryUsage"
+              }
+            }
+          },
+          "2023-04-02T00:00:00Z":{}
```
saltgen, line 1571: This one is showing up as a string on swagger UI.

```
"2023-04-02T00:00:00Z": "string"
```
Could you re-check by including following for the same?

```
        "type": "object",
        "additionalProperties": false
```

internal/types/kruizePayload/common.go
```diff
@@ -72,6 +74,26 @@ type RecommendationTerm struct {
 		Cost        RecommendationEngineObject `json:"cost,omitempty"`
 		Performance RecommendationEngineObject `json:"performance,omitempty"`
 	} `json:"recommendation_engines,omitempty"`
+	Plots Plot `json:"plots,omitempty"`
+}
+
+type Plot struct {
+	DataPoints int                  `json:"datapoints"`
```
saltgen, line 81:
```suggestion
	Plots *Plot `json:"plots,omitempty"`
}

type Plot struct {
	DataPoints int                  `json:"datapoints,omitempty"`
```

In case the env var; `plots=true` is not there for Kruize, ROS won't save any box plots json in the database.

kruize-clowdapp.yaml
```diff
@@ -101,6 +101,8 @@ objects:
             value: ${KRUIZE_CW_LOG_STREAM}
           - name: logging_cloudwatch_logLevel
             value: ${KRUIZE_CW_LOGGING_LEVEL}
+          - name: plots
+            value: "true"
```
saltgen, line 105: We'll need to grab the env var in kruize-clowdapp.

```
          - name: plots
            value: ${PLOTS_DATA}

- name: PLOTS_DATA
  value: "false"
```
The "false" should get overridden through deployment repo params.

saltgen (changes requested):
@upadhyeammit I don't think we should save plots data on ROS end if the flag is off for current Kruize image.
I hope you find these suggestions helpful

internal/types/kruizePayload/common.go
saltgen: Changes pushed

kruize-clowdapp.yaml
saltgen: Changes pushed

PR #186, opened 2024-02-12
Fixes #RHINENG-8204 - Check for required columns in CSV

patilsuraj767:

To detect if any of the column is missing.
If any column is missing pod goes into crashloop. 

- [ ] Yes
- [x] No

## Security Checklist :lock:

- [x] Bugfix
- [ ] New Feature
- [ ] Refactor
- [x] Unittests Added
- [ ] DRY code
- [ ] Dependency Added
- [ ] DB Migration Added

### DIFF
--- a/internal/utils/aggregator.go
+++ b/internal/utils/aggregator.go
@@ -1,6 +1,8 @@
 package utils
 
 import (
+	"fmt"
+	"reflect"
 	"strings"
 
 	"github.com/go-gota/gota/dataframe"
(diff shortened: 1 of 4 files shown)


internal/utils/aggregator.go
```diff
@@ -147,3 +155,51 @@ func determine_k8s_object_type(df dataframe.DataFrame) dataframe.DataFrame {
...
+		"memory_usage_container_min",
+		"memory_usage_container_max",
+		"memory_usage_container_sum",
+		"memory_rss_usage_container_avg",
+		"memory_rss_usage_container_min",
+		"memory_rss_usage_container_max",
+		"memory_rss_usage_container_sum",
+	}
+	cloumns_in_csv := df.Names()
+	if !reflect.DeepEqual(all_required_columns, cloumns_in_csv) {
```
upadhyeammit, line 201: Before doing the DeepEqual do you think we can just compare if the cloumns_in_csv is exactly having 37 number of items? if its not there then just error out? I don't know if this is already handled in DeepEqual, however I feel this may save sometime for us, considering we are doing this for every CSV?

Apart from that do you think we would ever need to know what are missing columns if in case of error?

upadhyeammit: Okay, looking further into it, it seems DeepEqual checks the length: https://github.com/goccy/go-reflect/blob/4646ad15ec8afb955a3bfb6af969f007f169f383/reflect.go#L364

patilsuraj767: Looks like it is already handled in DeepEqual - https://github.com/golang/go/blob/master/src/reflect/deepequal.go#L102-#L103

patilsuraj767:
Thanks for the review @upadhyeammit. Merging the PR.

PR #187, opened 2024-02-15
Tuning kafka consumer config

patilsuraj767:

There are kafka metrics gaps observed in grafana. Possible cause for this is rebalancing which occurs in kafka consumer groups. 

Below are the reference link for the config params `session.timeout.ms` and `heartbeat.interval.ms`.
https://kafka.apache.org/34/documentation.html#consumerconfigs_heartbeat.interval.ms
https://kafka.apache.org/34/documentation.html#consumerconfigs_session.timeout.ms

- [ ] Yes
- [x] No

## Security Checklist :lock:

- [ ] Bugfix
- [ ] New Feature
- [ ] Refactor
- [ ] Unittests Added
- [ ] DRY code
- [ ] Dependency Added
- [ ] DB Migration Added

### DIFF
--- a/internal/kafka/consumer.go
+++ b/internal/kafka/consumer.go
@@ -39,6 +39,8 @@ func StartConsumer(kafka_topic string, handler func(msg *kafka.Message, consumer
 			"enable.auto.commit":       auto_commit,
 			"go.logs.channel.enable":   true,
 			"allow.auto.create.topics": true,
+			"session.timeout.ms":       120000,
+			"heartbeat.interval.ms":    30000,
 		}
 
 		// As per librdkafka doc - https://github.com/confluentinc/librdkafka/blob/master/CONFIGURATION.md?plain=1#L73


saltgen (approved):

upadhyeammit:
Hello!

Below is my understanding, correct me if I am wrong!

The default session.timeout.ms is 45 seconds and heartbeat.interval.ms is 3 seconds. This does mean before re balancing consumers tries 15 times and that is spread across 45 seconds.
The updated parameters are 120 seconds for  session.timeout.ms and 30 seconds for heartbeat.interval.ms, which means this would try 4 times spread across 120 seconds.

Considering these it seems change has below questions,
1. If the consumer is not available or frequently not available for 45 seconds in production then I don't think this is good sign, and we should not increase it to 120?
2. Waiting for 30 seconds for each heartbeat seems slower to me, as for that amount of time all Kafka consumers stop processing the messages? The consumer should join as early as possible?

patilsuraj767:
Thanks for the review @saltgen. I have also tested in EE and it is working without any issue.

patilsuraj767:
> 1. If the consumer is not available or frequently not available for 45 seconds in production then I don't think this is good sign, and we should not increase it to 120?

The meaning of consumer not available here means that consumer is processing the earlier message and it is taking more than 45 second to process it and I think this is expectable. Hence I increased the timeout to 120 seconds.
  
> 2. Waiting for 30 seconds for each heartbeat seems slower to me, as for that amount of time all Kafka consumers stop processing the messages? The consumer should join as early as possible?

Kafka consumer only gets disconnected(rebalance occur) if next message is not poll from the topic within `session.timeout` or if there is no single heartbeat send by the consumer to broker within the `session.timeout`. Other than this consumer stops if there is any issue on broker side. 
Here it is a trade-off. 3 second heartbeat and 45 seconds timeout is good for the application which are picking and processing the msgs quickly. As a nature of our application we cannot increase the processing time. Hence we will need to do 30 seconds compromise.

upadhyeammit:
For understanding purpose:

~~~
I think our report processor pods won't ever take 45 seconds to process a message? however the poller may need 45 secs or more?
=>I think it is possible processor is taking more than 45 seconds. Because for a large cluster there can be many deployments in single CSV. and to add all the metrics to ROS DB, to create the experiments in the Kruize and then to hit kruize's update_metrics API these 3 things may take 45seconds or more.
~~~

With this understanding we are proceeding with updated parameters.

upadhyeammit (approved):

patilsuraj767:
Thanks for the review @upadhyeammit. Merging the PR.

PR #189, opened 2024-02-20
Fixes #RHINENG-8344 - Fix validation of incoming reports

patilsuraj767:

Processor service crashes with the below error. 
```
panic: runtime error: makeslice: len out of range

goroutine 1 [running]:
github.com/go-gota/gota/dataframe.DataFrame.Rapply({{0xc0000f6a80, 0x25, 0x25}, 0x25, 0x0, {0x0, 0x0}}, 0xc001dea850)
	/opt/app-root/src/go/pkg/mod/github.com/go-gota/gota@v0.12.0/dataframe/dataframe.go:899 +0x785
github.com/redhatinsights/ros-ocp-backend/internal/utils.determine_k8s_object_type({{0xc0000f6a80, 0x25, 0x25}, 0x25, 0x0, {0x0, 0x0}})
	/go/src/app/internal/utils/aggregator.go:140 +0x6d0
github.com/redhatinsights/ros-ocp-backend/internal/utils.Aggregate_data({{0xc0000baa80, 0x25, 0x25}, 0x25, 0x1ae, {0x0, 0x0}})
	/go/src/app/internal/utils/aggregator.go:23 +0x12b
github.com/redhatinsights/ros-ocp-backend/internal/services.ProcessReport(0xc0018dab00, 0x4?)
	/go/src/app/internal/services/report_processor.go:75 +0x73d
github.com/redhatinsights/ros-ocp-backend/internal/kafka.StartConsumer({0xc0003daef0, 0xf}, 0x1894728, {0x0, 0x0, 0x0?})
	/go/src/app/internal/kafka/consumer.go:77 +0xa89
github.com/redhatinsights/ros-ocp-backend/cmd.glob..func6(0x25e53a0?, {0x181057b?, 0x0?, 0x0?})
	/go/src/app/cmd/start.go:25 +0x91
github.com/spf13/cobra.(*Command).execute(0x25e53a0, {0x26a29e0, 0x0, 0x0})
	/opt/app-root/src/go/pkg/mod/github.com/spf13/cobra@v1.6.1/command.go:920 +0x847
github.com/spf13/cobra.(*Command).ExecuteC(0x25e4de0)
	/opt/app-root/src/go/pkg/mod/github.com/spf13/cobra@v1.6.1/command.go:1044 +0x3bd
github.com/spf13/cobra.(*Command).Execute(...)
	/opt/app-root/src/go/pkg/mod/github.com/spf13/cobra@v1.6.1/command.go:968
github.com/redhatinsights/ros-ocp-backend/cmd.Execute()
	/go/src/app/cmd/root.go:16 +0x25
main.main()
	/go/src/app/rosocp.go:6 +0x17

```
This happens because we are trying to modify empty dataframe. 
Validation must happen before processing any CSV file. Also after filtering we should make sure that dataframe is not empty. 

- [x] Bugfix
- [ ] New Feature
- [ ] Refactor
- [ ] Unittests Added
- [ ] DRY code
- [ ] Dependency Added
- [ ] DB Migration Added

## Additional :mega:

### DIFF
--- a/internal/utils/aggregator.go
+++ b/internal/utils/aggregator.go
@@ -28,8 +28,7 @@ func Aggregate_data(df dataframe.DataFrame) (dataframe.DataFrame, error) {
 	}
 
 	if df.Nrow() == 0 {
-		log.Error("No valid records present in CSV to process further.")
-		return df, nil
+		return df, fmt.Errorf("no valid records present in CSV to process further")
 	}
 
 	df = determine_k8s_object_type(df)
@@ -102,8 +101,8 @@ func filter_valid_csv_records(main_df dataframe.DataFrame) (dataframe.DataFrame,
 		dataframe.F{Colname: "cpu_usage_container_avg", Comparator: series.GreaterEq, Comparando: 0},
 		dataframe.F{Colname: "owner_kind", Comparator: series.Neq, Comparando: ""},
 		dataframe.F{Colname: "owner_name", Comparator: series.Neq, Comparando: ""},
-		dataframe.F{Colname: "workload", Comparator: series.Neq, Comparando: ""},
-		dataframe.F{Colname: "workload_type", Comparator: series.Neq, Comparando: ""},
+		dataframe.F{Colname: "owner_kind", Comparator: series.Neq, Comparando: "<none>"},
+		dataframe.F{Colname: "owner_name", Comparator: series.Neq, Comparando: "<none>"},
 	)
 
 	no_of_dropped_records := main_df.Nrow() - df.Nrow()
@@ -139,9 +138,9 @@ func determine_k8s_object_type(df dataframe.DataFrame) dataframe.DataFrame {
 		owner_kind := s.Elem(index_of_owner_kind).String()
 		workload := s.Elem(index_of_workload).String()
 		workload_type := s.Elem(index_of_workload_type).String()
-		if strings.ToLower(owner_kind) == string(w.Replicaset) && workload == "<none>" {
+		if strings.ToLower(owner_kind) == string(w.Replicaset) && (workload == "<none>" || workload == "") {
 			return series.Strings([]string{string(w.Replicaset), owner_name})
-		} else if strings.ToLower(owner_kind) == string(w.Replicationcontroller) && workload == "<none>" {
+		} else if strings.ToLower(owner_kind) == string(w.Replicationcontroller) && (workload == "<none>" || workload == "") {
 			return series.Strings([]string{string(w.Replicationcontroller), owner_name})
 		} else {
 			return series.Strings([]string{workload_type, workload})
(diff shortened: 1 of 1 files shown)


saltgen (commented):
@patilsuraj767 IMO it would be good to maintain a set of operator versions for ROS OCP, thoughts?

internal/utils/aggregator.go
```diff
@@ -99,6 +100,10 @@ func filter_valid_csv_records(main_df dataframe.DataFrame) (dataframe.DataFrame,
 		dataframe.F{Colname: "cpu_usage_container_max", Comparator: series.GreaterEq, Comparando: 0},
 		dataframe.F{Colname: "cpu_usage_container_min", Comparator: series.GreaterEq, Comparando: 0},
 		dataframe.F{Colname: "cpu_usage_container_avg", Comparator: series.GreaterEq, Comparando: 0},
+		dataframe.F{Colname: "owner_kind", Comparator: series.Neq, Comparando: ""},
+		dataframe.F{Colname: "owner_name", Comparator: series.Neq, Comparando: ""},
```
saltgen, line 104: @patilsuraj767  Do we need to check for both of these as well? Could you share some sample values?

patilsuraj767: Yes, you can check row 104 in our sample CSV - https://github.com/RedHatInsights/ros-ocp-backend/blob/main/scripts/samples/ros-ocp-usage.csv#L104

internal/utils/aggregator.go
```diff
@@ -33,9 +28,15 @@ func Aggregate_data(df dataframe.DataFrame) (dataframe.DataFrame, error) {
 	}
 
 	if df.Nrow() == 0 {
+		log.Error("No valid records present in CSV to process further.")
 		return df, nil
```
upadhyeammit, line 32: I think here we should return the err instead of nil? because in dependent func we are checking if there is nil, if not nil then we end up processing the df further?

https://github.com/RedHatInsights/ros-ocp-backend/blob/main/internal/services/report_processor.go#L75

patilsuraj767: Yes, you are right. 
Thanks. I will make the changes.

internal/utils/aggregator.go
patilsuraj767: BTW I thought on this again and I realised that we need `owner_kind` and `owner_name` both but we do not need `workload` and `workload_type` so removing it.

saltgen: Meant to ask if these values are missing as well along with workload and workload_type, as per the cost card

saltgen: Meant to ask if these values are missing as well along with workload and workload_type, as per the cost card

upadhyeammit (approved):

saltgen (approved):
LGTM

patilsuraj767:
Thanks for the review @upadhyeammit and @saltgen. Merging the PR.

PR #191, opened 2024-02-21
Use different Prometheus ports for every server locally

upadhyeammit:

The report-processor, recommendation-poller and api-server starts prometheus server, however the port for prometheus is hardcoded 5005 in the config. This change ties different PROMETHEUS_PORT for cmds in Makefile and then viper's AutomaticEnv() and GetString() is used to consume it.

With this change one can run all of the report-processor, recommendation-poller and api-server at a time without stopping any of them.

- [ ] Yes
- [ ] No

## Security Checklist :lock:

- [ ] Bugfix
- [ ] New Feature
- [ ] Refactor
- [ ] Unittests Added
- [ ] DRY code
- [ ] Dependency Added
- [ ] DB Migration Added

## Additional :mega:

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -60,15 +60,15 @@ db-migrate:
 
 .PHONY: run-processor
 run-processor:
-	go run rosocp.go start processor
+	PROMETHEUS_PORT=5005 go run rosocp.go start processor
 
 .PHONY: run-recommender
 run-recommendation-poller:
-	go run rosocp.go start recommendation-poller
+	PROMETHEUS_PORT=5006 go run rosocp.go start recommendation-poller
 
 .PHONY: run-api-server
 run-api-server:
-	go run rosocp.go start api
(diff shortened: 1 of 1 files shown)


internal/config/config.go
```diff
@@ -154,7 +155,7 @@ func initConfig() {
 		viper.SetDefault("RBAC_ENABLE", false)
 
 		// prometheus config
-		viper.SetDefault("PROMETHEUS_PORT", "5005")
+		viper.GetString("PROMETHEUS_PORT")
```
patilsuraj767, line 158: Can we keep some default value for this as it was before? Some OS allow binding multiple service on same port.

patilsuraj767 (changes requested):

patilsuraj767 (approved):
Thanks @upadhyeammit LGTM 👍

PR #194, opened 2024-03-04
Fixes #RHINENG-7907 - use monitoring_end_time with start_date

upadhyeammit:

On first of the month api returns nothing even though recommendations are stored. This is because monitoring_start_time for that recommendation is older than 1st of the current month. To fix this the filter has been changed to depend on monitoring_end_time. This does mean if we receive data partially between 28,29,30 or 31 to 1st then we show the recommendation if its there in OCP database.

With this I have also updated the logic to show error upon invalid/bad start_data and end_date instead of just logging error in service. This does mean if invalid date is provided we don't try to query with that to database, so fewer queries when we know param(s) are bad!
 
- [ ] Yes
- [ ] No

## Security Checklist :lock:

- [x] Bugfix
- [ ] New Feature
- [ ] Refactor
- [ ] Unittests Added
- [ ] DRY code
- [ ] Dependency Added
- [ ] DB Migration Added

## Additional :mega:

### DIFF
--- a/internal/api/handlers.go
+++ b/internal/api/handlers.go
@@ -16,8 +16,13 @@ func GetRecommendationSetList(c echo.Context) error {
 	XRHID := c.Get("Identity").(identity.XRHID)
 	OrgID := XRHID.Identity.OrgID
 	user_permissions := get_user_permissions(c)
+	var orderHow string
+	var orderBy string
+	// Default values
+	var limit int = 10
+	var offset int = 0
 
-	orderBy := c.QueryParam("order_by")
+	orderBy = c.QueryParam("order_by")
 	if orderBy != "" {
 		var orderByOptions = map[string]string{
 			"cluster":       "clusters.cluster_alias",
@@ -37,7 +42,7 @@ func GetRecommendationSetList(c echo.Context) error {
 		orderBy = "clusters.last_reported_at"
 	}
 
-	orderHow := c.QueryParam("order_how")
+	orderHow = c.QueryParam("order_how")
 	if orderHow != "" {
 		orderHowUpper := strings.ToUpper(orderHow)
 		if (orderHowUpper != "ASC") && (orderHowUpper != "DESC") {
@@ -51,7 +56,6 @@ func GetRecommendationSetList(c echo.Context) error {
 	orderQuery := orderBy + " " + orderHow
 
 	limitStr := c.QueryParam("limit")
-	limit := 10 // default value
 	if limitStr != "" {
 		limitInt, err := strconv.Atoi(limitStr)
 		if err == nil {
@@ -60,7 +64,7 @@ func GetRecommendationSetList(c echo.Context) error {
 	}
 
 	offsetStr := c.QueryParam("offset")
-	offset := 0 // default value
+
 	if offsetStr != "" {
 		offsetInt, err := strconv.Atoi(offsetStr)
 		if err == nil {
@@ -68,7 +72,10 @@ func GetRecommendationSetList(c echo.Context) error {
 		}
 	}
 
-	queryParams := MapQueryParameters(c)
+	queryParams, err := MapQueryParameters(c)
+	if err != nil {
+		return c.JSON(http.StatusBadRequest, echo.Map{"status": "error", "message": err.Error()})
+	}
 	recommendationSet := model.RecommendationSet{}
 	recommendationSets, count, error := recommendationSet.GetRecommendationSets(OrgID, orderQuery, limit, offset, queryParams, user_permissions)
 	if error != nil {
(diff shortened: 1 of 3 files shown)


upadhyeammit:
I wish to add some tests around this, however this is ready for review!

internal/api/handlers.go
```diff
@@ -68,7 +68,10 @@ func GetRecommendationSetList(c echo.Context) error {
 		}
 	}
 
-	queryParams := MapQueryParameters(c)
+	queryParams, err := MapQueryParameters(c)
+	if err != nil {
+		return c.JSON(http.StatusBadRequest, err)
```
saltgen, line 73: Unsure, if we should send this error as JSON, could you provide a sample?

saltgen (changes requested):

internal/api/handlers.go
upadhyeammit: We are very perticular with respect to converting the given start/end date to Time format, and so we are also raising the error in backend about we are unable to convert it. However at the user end if the problem is in format of the date then we just process further, however the results will be incorrect? So for sample we can use any date which is not in YYYY-MM-DD format and api works, which should not. With this change it shows error as below,
~~~
* Connection #0 to host localhost left intact
{
    "message": "parsing time \"1970-1-01\" as \"2006-01-02\": cannot parse \"1-01\" as \"01\"",
    "status": "error"
}
~~~

saltgen: We could send this error as json to user, however IMO we should use a standard date probably `1970-01-01` for comparison.

saltgen: Additionally, could we type assert the other params as well?

upadhyeammit: > We could send this error as json to user, however IMO we should use a standard date probably `1970-01-01` for comparison in `start_date` and `end_date`.

We cant use 1970-01-01 as time.Parse function does not accept it as per RFC3339. So not doing this change

upadhyeammit: > Additionally, could we type assert the other params as well?
> 
> Instead of implicit type casting i.e. using `:=` I think we should declare type explicitly and then grab the values using `c.QueryParams()`

Done!

saltgen (changes requested):

internal/api/utils.go
```diff
@@ -83,74 +85,82 @@ func CollectionResponse(collection []interface{}, req *http.Request, count, limi
...
+func MapQueryParameters(c echo.Context) (map[string][]string, error) {
 	log := logging.GetLogger()
 	queryParams := make(map[string][]string)
+	var startDate time.Time
+	var endDate time.Time
+	var clusters []string
+	var projects []string
+	var workloadNames []string
+	var workloadTypes []string
+	var containers []string
```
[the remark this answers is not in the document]
saltgen: Discussed offline

saltgen (approved):
Thanks for the changes @upadhyeammit 🙌🏼

saltgen:
> I wish to add some tests around this, however this is ready for review!

Let me know when this MR is good to merge, thanks.

internal/api/utils_test.go
```diff
@@ -0,0 +1,83 @@
...
+			qinputs: map[string]string{"start_date": "2023-03-23", "end_date": "2023-03-24"},
+			qoutputs: map[string][]string{
+				"DATE(recommendation_sets.monitoring_end_time) <= ?": {"2023-03-24"},
+				"DATE(recommendation_sets.monitoring_end_time) >= ?": {"2023-03-23"},
+			},
+			errmsg: `The recommendation_sets.monitoring_end_time should be less than or equal to end date!
+				The recommendation_sets.monitoring_end_time should be greater than or equal to start date!`,
+		},
+		{
+			name:    "When start date and end date are not provided",
```
saltgen, line 58: Regarding the first two test cases default values should kick in, additionally I don't see a test case where-in we are checking if start date is compared against monitoring_end_time instead of monitoring_start_time.
Are we trying to gather that from the json error response?

saltgen (changes requested):

internal/api/utils_test.go
saltgen: @upadhyeammit As discussed, we will need a setup and teardown system to test the change in this PR, that said we can keep the below test in the highlighted section.

`name: "When start date and end date are provided"`

saltgen (changes requested):

internal/api/utils_test.go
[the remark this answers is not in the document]
upadhyeammit: okay, however the test which we agreed to keep is only for func MapQueryParameters and not for testing the api completely? Hence it makes me feel to keep as utils_test.go?

saltgen (approved):

PR #198, opened 2024-03-28
RHINENG-9051  CPU Format value nullable

saltgen:

CPU format can be string or null as per business logic.

- [ ] Yes
- [ ] No

## Security Checklist :lock:

- [ ] Bugfix
- [ ] New Feature
- [ ] Refactor
- [ ] Unittests Added
- [ ] DRY code
- [ ] Dependency Added
- [ ] DB Migration Added

## Additional :mega:

### DIFF
--- a/openapi.json
+++ b/openapi.json
@@ -294,7 +294,8 @@
                           },
                           "format": {
                             "type": "string",
-                            "example": null
+                            "example": null,
+                            "nullable": true
                           }
                         }
                       },
@@ -325,7 +326,8 @@
                           },
                           "format": {
                             "type": "string",
-                            "example": null
+                            "example": null,
+                            "nullable": true
                           }
                         }
                       },
@@ -551,7 +553,8 @@
                       },
                       "format": {
                         "type": "string",
-                        "example": null
+                        "example": null,
+                        "nullable": true
                       }
                     }
                   },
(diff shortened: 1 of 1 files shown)


upadhyeammit:
Considering the format can be null however still there could be valid integer values for CPU, I just feel the recommendation we generate and pass on to user maybe not 100% legit if,
1. We never received the format from cost operator?
2. We received it and passed it on to Kruize however we received null format from Kruize?

saltgen:
> Considering the format can be null however still there could be valid integer values for CPU, I just feel the recommendation we generate and pass on to user maybe not 100% legit if,
> 
> 1. We never received the format from cost operator?
> 2. We received it and passed it on to Kruize however we received null format from Kruize?
> 
> Edit: And this can also happen with Memory?

The format that we receive from the cost operator is implicit and not exactly mentioned.

Additionally these format changes are updated by the API helper method, so if any issues surface it would show up before we reach this stage.

Format won't be null for memory as well, at this stage.

saltgen:
@yash2189 Changes awaiting your ack

P.S. Let's target all other issues (if any) w.r.t openapi spec and goatee tests here.

upadhyeammit (approved):

yash2189:
I am a bit confused in this [line](https://github.com/RedHatInsights/ros-ocp-backend/blob/5ef7541371c6ae53bf9f71f663d8c693402419c0/openapi.json#L355) of openapi spec where we mention that notification code type is string but in example we have provided 111101 without any quotes
In other places where code is defined we have properly provided example as per given format

PR #205, opened 2024-04-12
RHINENG-5297 Notification filtering

saltgen:

API response should include a predetermined list of notifications.

- [ ] Yes
- [x] No

## Security Checklist :lock:

- [ ] Bugfix
- [x] New Feature
- [ ] Refactor
- [ ] Unittests Added
- [ ] DRY code
- [ ] Dependency Added
- [ ] DB Migration Added

### Logs
```
WARN[0002]~/ros-ocp-backend/internal/api/utils.go:386 github.com/redhatinsights/ros-ocp-backend/internal/api.filterNotifications.func1() dropped notification 111101 dropped from recommendation ID: 3fdaae29-57f9-4782-b7be-646ef45dad68; cluster ID: 222  service=rosocp
```

Note: There seems to be too many logs per recommendation

Above issue is now addressed.

```
WARN[0019]~/ros-ocp-backend/internal/api/utils.go:396 github.com/redhatinsights/ros-ocp-backend/internal/api.filterNotifications() 111101, 112101, 112102, 120001, 120001 dropped from recommendation ID: 66d1445b-976a-4e8b-9013-56111973044f; cluster ID: 222  service=rosocp

```

### DIFF
--- a/internal/api/utils.go
+++ b/internal/api/utils.go
@@ -353,6 +353,8 @@ func transformComponentUnits(recommendationJSON map[string]interface{}) map[stri
 
 func filterNotifications(recommendationID string, clusterUUID string, recommendationJSON map[string]interface{}) map[string]interface{} {
 
+	var droppedNotifications []string
+
 	deleteNotificationObject := func(recommendationSection map[string]interface{}) {
 		notificationObject, ok := recommendationSection["notifications"].(map[string]interface{})
 		if ok {
@@ -360,7 +362,7 @@ func filterNotifications(recommendationID string, clusterUUID string, recommenda
 				_, found := NotificationsToShow[key]
 				if !found {
 					delete(recommendationSection, "notifications")
-					log.Warnf("dropped notification %s dropped from recommendation ID: %s; cluster ID: %s", key, recommendationID, clusterUUID)
+					droppedNotifications = append(droppedNotifications, key)
 				}
 			}
 		}
@@ -393,6 +395,9 @@ func filterNotifications(recommendationID string, clusterUUID string, recommenda
 			}
 		}
 	}
+	droppedNotificationsString := strings.Join(droppedNotifications, ", ")
+	log.Warnf("%s dropped from recommendation ID: %s; cluster ID: %s", droppedNotificationsString, recommendationID, clusterUUID)
+
 	return recommendationJSON
 }
 
--- a/internal/api/utils.go
+++ b/internal/api/utils.go
@@ -16,34 +16,6 @@ import (
 	"github.com/redhatinsights/ros-ocp-backend/internal/logging"
 )
 
-const timeLayout = "2006-01-02"
-
-var NotificationsToShow = map[string]string{
-	"323004": "NOTICE",
-	"323005": "NOTICE",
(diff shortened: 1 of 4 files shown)


internal/api/utils.go
```diff
@@ -16,6 +16,13 @@ import (
 	"github.com/redhatinsights/ros-ocp-backend/internal/logging"
 )
 
+var NotificationsToShow = map[string]string{
```
upadhyeammit, line 19: This just feels does not belong to utils, as per what goes where /ros-ocp-backend/blob/main/internal/types/kruizePayload/common.go feels better way to keep this?

internal/api/utils.go
```diff
@@ -370,5 +371,64 @@ func TransformComponentUnits(jsonData datatypes.JSON) map[string]interface{} {
...
+			for key := range notificationObject {
+				_, found := NotificationsToShow[key]
+				if !found {
+					delete(recommendationSection, "notifications")
+					log.Warnf("dropped notification %s dropped from recommendation ID: %s; cluster ID: %s", key, recommendationID, clusterUUID)
+				}
+			}
+		}
+
+	}
```
upadhyeammit, line 391: I would have preferred separate named func, however I can also understand this block of code is required by filterNotifications only! At the same time we can have multiple utils as per API of functionality. So I leave it upto you to keep it as is or create new func.

internal/api/utils.go
```diff
@@ -370,5 +371,64 @@ func TransformComponentUnits(jsonData datatypes.JSON) map[string]interface{} {
...
+		log.Error("recommendation data not found in JSON")
+		return recommendationJSON
+	}
+
+	for _, term := range []string{"short_term", "medium_term", "long_term"} {
+		levelThree, ok := recommendationTerms[term].(map[string]interface{})
+		if ok {
+			deleteNotificationObject(levelThree)
+		}
+		recommendationEngineObject, ok := levelThree["recommendation_engines"].(map[string]interface{})
```
upadhyeammit, line 409: If I see the [kruize docs](https://github.com/kruize/autotune/blob/remote_monitoring/design/NotificationCodes.md) then all of the recommendation codes are of engine level, so should we not just remove that from engine level?

saltgen: I think you are talking about `NotificationsToShow` codes

internal/api/utils.go
saltgen: Let's convert this to a named function if it gets re-used later

internal/api/utils.go
saltgen: I have moved the constant and some other structs to a dedicated `common.go` file for the api package

saltgen: I have moved the constant and some other structs to a dedicated `common.go` file for the api package

saltgen:
@upadhyeammit Any thoughts on the Additional section, regarding logs?

kgaikwad:
Hello @saltgen  could you please add ticket RHINENG-5297 as this is ROS OCP side ticket and RHINENG-9241 is Epic ticket includes Kruize side change too.

upadhyeammit (approved):

PR #206, opened 2024-04-15
Fixes #RHINENG-5352 - Implementing delete API of kruize

patilsuraj767:

To delete the kruize side data when source is deleted from console.redhat.com. 

- [ ] Yes
- [x] No

## Security Checklist :lock:

- [ ] Bugfix
- [x] New Feature
- [ ] Refactor
- [ ] Unittests Added
- [ ] DRY code
- [ ] Dependency Added
- [ ] DB Migration Added

### DIFF
--- a/internal/services/housekeeper/sourcesCleaner.go
+++ b/internal/services/housekeeper/sourcesCleaner.go
@@ -2,6 +2,7 @@ package housekeeper
 
 import (
 	"encoding/json"
+	"fmt"
 	"os"
 	"strconv"
 
@@ -9,10 +10,12 @@ import (
 	"github.com/labstack/gommon/log"
 
 	"github.com/redhatinsights/ros-ocp-backend/internal/config"
+	database "github.com/redhatinsights/ros-ocp-backend/internal/db"
 	"github.com/redhatinsights/ros-ocp-backend/internal/kafka"
 	"github.com/redhatinsights/ros-ocp-backend/internal/logging"
 	"github.com/redhatinsights/ros-ocp-backend/internal/model"
 	"github.com/redhatinsights/ros-ocp-backend/internal/types"
+	"github.com/redhatinsights/ros-ocp-backend/internal/utils/kruize"
 	"github.com/redhatinsights/ros-ocp-backend/internal/utils/sources"
 )
 
@@ -32,6 +35,7 @@ func StartSourcesListenerService() {
 }
 
 func sourcesListener(msg *k.Message, _ *k.Consumer) {
+	db := database.GetDB()
 	headers := msg.Headers
 	for _, v := range headers {
 		if v.Key == "event_type" && string(v.Value) == "Application.destroy" {
@@ -45,9 +49,19 @@ func sourcesListener(msg *k.Message, _ *k.Consumer) {
 				return
 			}
 			if data.Application_type_id == cost_app_id {
-				cluster := model.Cluster{
-					SourceId: strconv.Itoa(data.Source_id),
+				var cluster model.Cluster
+				fmt.Println(data.Source_id)
+				db.Where("source_id = ?", strconv.Itoa(data.Source_id)).First(&cluster)
+				workloads, err := model.GetWorkloadsByClusterID(cluster.ID)
+				if err != nil {
+					log.Errorf("unable to get workloads for cluster: %v. Error: %v", cluster, err)
+					return
 				}
+
+				for _, workload := range workloads {
+					kruize.Delete_experiment_from_kruize(workload.ExperimentName)
+				}
+
 				if err := cluster.DeleteCluster(); err != nil {
 					log.Errorf("unable to delete record from clusters table: %v. Error: %v", cluster, err)
 				} else {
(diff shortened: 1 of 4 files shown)


internal/services/housekeeper/sourcesCleaner.go
```diff
@@ -32,6 +34,7 @@ func StartSourcesListenerService() {
 }
 
 func sourcesListener(msg *k.Message, _ *k.Consumer) {
+	db := database.GetDB()
 	headers := msg.Headers
 	for _, v := range headers {
 		if v.Key == "event_type" && string(v.Value) == "Application.destroy" {
```
upadhyeammit, line 40: The "Application.destroy" happens after successfully deleting the app and not just triggering the destroy/delete action?

internal/utils/kruize/kruize_api.go
```diff
@@ -196,3 +196,32 @@ func Is_valid_recommendation(recommendation kruizePayload.Recommendation, experi
...
+func Delete_experiment_from_kruize(experiment_name string) {
+	url := cfg.KruizeUrl + "/createExperiment"
+	data := []map[string]string{
+		{"experiment_name": experiment_name},
+	}
+	payload, err := json.Marshal(data)
+	if err != nil {
+		log.Error("unable to marshal payload to json: ", err)
+		return
+	}
```
upadhyeammit, line 209: Is this needed? considering the `data` and `experiment_name` are statically defined?
and if we need this, then we should also increase counter and message that we failed to delete exp because we have read the kafka message and in next iteration we will delete next one?,
~~~
kruizeAPIException.WithLabelValues("/deleteExperiment").Inc()
log.Error("error Occured while deleting experiment: ", err)
~~~

internal/utils/kruize/kruize_api.go
```diff
@@ -196,3 +196,32 @@ func Is_valid_recommendation(recommendation kruizePayload.Recommendation, experi
...
+
+	client := &http.Client{}
+	req, err := http.NewRequest("DELETE", url, bytes.NewBuffer(payload))
+	if err != nil {
+		log.Error("unable to create request object ", err)
```
upadhyeammit, line 214: same here, we should increment counter and err about deletion,
~~~
kruizeAPIException.WithLabelValues("/deleteExperiment").Inc()
log.Error("error Occured while deleting experiment: ", err)
~~~

internal/utils/kruize/kruize_api.go
```diff
@@ -196,3 +196,32 @@ func Is_valid_recommendation(recommendation kruizePayload.Recommendation, experi
...
+		return
+	}
+	res, err := client.Do(req)
+	if err != nil {
+		kruizeAPIException.WithLabelValues("/deleteExperiment").Inc()
+		log.Error("error Occured while deleting experiment: ", err)
+		return
+	}
+	defer res.Body.Close()
+	if res.StatusCode == 201 {
```
upadhyeammit, line 224: The generic search says,
~~~
If the item is was successfully deleted, return 200 (the resource exists as of the time of the request, and the action was executed). If it's being queued for execution, return 202. If the resource has already been deleted, return 410. If it otherwise doesn't exist, return 404.
~~~

If I see the kruize docs then they are returning 201 however are there chances of 200/202/410/404?

internal/services/housekeeper/sourcesCleaner.go
patilsuraj767: Not sure if I get you question but `Application.destroy` event is triggered when cluster is removed from the sources in consoledot.

internal/utils/kruize/kruize_api.go
patilsuraj767: Agree, It is not needed. I will remove it.

internal/utils/kruize/kruize_api.go
patilsuraj767: I will add the counter and change the error message.

internal/utils/kruize/kruize_api.go
patilsuraj767: Apart from `201` kruize does send `400` if it fails to delete the experiment. I will add the else block to increase prometheus counter and to log error.

internal/utils/kruize/kruize_api.go
```diff
@@ -196,3 +196,32 @@ func Is_valid_recommendation(recommendation kruizePayload.Recommendation, experi
...
+	}
+	payload, _ := json.Marshal(data)
+
+	client := &http.Client{}
+	req, err := http.NewRequest("DELETE", url, bytes.NewBuffer(payload))
+	if err != nil {
+		kruizeAPIException.WithLabelValues("/deleteExperiment").Inc()
+		log.Error("error occured while deleting experiment: ", err)
+		return
+	}
```
upadhyeammit, line 213: Can this be de-duplicated? maybe a separate small func can do the job? or if you wish to keep that inside `Delete_experiment_from_kruize` then var=unamed func and reuse it thrice?

patilsuraj767: Done

upadhyeammit (approved):

patilsuraj767:
Thanks for the review @upadhyeammit

PR #207, opened 2024-04-18
Add new flags for kruize local and logging

khansaad:
## PR Title :boom:

This PR adds new flags in the `kruize-clowdapp.yaml`

To enable `kruize-local` use case and for logging request and response for debugging purpose.

- [ ] Yes
- [x] No

## Security Checklist :lock:

  - If yes what is the version no: 0.0.21
     Yet to be released
- [ ] Bugfix
- [ ] New Feature
- [ ] Refactor
- [ ] Unittests Added
- [ ] DRY code
- [ ] Dependency Added
- [ ] DB Migration Added

## Additional :mega:


kruize-clowdapp.yaml
```diff
@@ -73,6 +73,10 @@ objects:
             value: "prometheus-k8s"
           - name: savetodb
             value: "true"
+          - name: local
+            value: "false"
+          - name: logAllHttpReqAndResp
+            value: "true"
```
saltgen, line 79: @khansaad "false" for `logAllHttpReqAndResp` makes sense here as well, we can override it on EE when required, also thoughts on switching to upper case?

saltgen (changes requested):

kruize-clowdapp.yaml
khansaad: @saltgen We must debug the reasons for the failure of the `updateRecommendation` calls. This flag needs to be set to `true` until the issue is resolved.

Switched the variable to upper case as suggested.

kruize-clowdapp.yaml
```diff
@@ -130,6 +134,10 @@ objects:
               value: "prometheus-k8s"
             - name: savetodb
               value: "true"
+            - name: local
+              value: "false"
+            - name: LOG_ALL_HTTP_REQ_AND_RESPONSE
+              value: "true"
```
saltgen, line 140: @khansaad I don't think we need these vars here, i.e. for the type job.

kruize-clowdapp.yaml
```diff
@@ -183,6 +191,10 @@ objects:
...
```
saltgen, line 197: Ditto

saltgen (changes requested):

kruize-clowdapp.yaml
saltgen: Ah okay, it seemed like `logAllHttpReqAndResp` was for kruize local.

kruize-clowdapp.yaml
khansaad: Updated now.

saltgen (approved):
Thanks @khansaad LGTM!

PR #209, opened 2024-04-19
Fixes #RHINENG-9588 - Remove unique constraint from rh_accounts

patilsuraj767:

To support anemic customers where account number is not present.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -9,8 +9,8 @@ else
     b64_identity=$(shell echo '${identity}' | base64 -w 0 -)
 endif
 
-ros_ocp_msg='{"request_id": "uuid1234", "b64_identity": "test", "metadata": {"account": "2234", "org_id": "3340851", "source_id": "111", "cluster_uuid": "222", "cluster_alias": "name222"}, "files": ["http://localhost:8888/ros-ocp-usage.csv"]}'
-ros_ocp_msg_24Hrs='{"request_id": "uuid1234", "b64_identity": "test", "metadata": {"account": "2234", "org_id": "3340851", "source_id": "111", "cluster_uuid": "222", "cluster_alias": "name222"}, "files": ["http://localhost:8888/ros-ocp-usage-24Hrs.csv"]}'
+ros_ocp_msg='{"request_id": "uuid1234", "b64_identity": "test", "metadata": {"org_id": "3340851", "source_id": "111", "cluster_uuid": "222", "cluster_alias": "name222"}, "files": ["http://localhost:8888/ros-ocp-usage.csv"]}'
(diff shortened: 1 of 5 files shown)


internal/model/rh_account.go
```diff
@@ -6,7 +6,7 @@ import (
 
 type RHAccount struct {
 	ID      uint   `gorm:"primaryKey;not null;autoIncrement"`
-	Account string `gorm:"type:text;unique"`
+	Account string `gorm:"type:text;"`
```
upadhyeammit, line 9: I just confirmed we don't consider both account and orgid while looking up for the rh_accounts, I can see we lookup based on orgId, so we are good even when account is sometimes empty and sometimes having some value,
~~~
rh_account.go:	
result := db.Where("org_id = ?", r.OrgId).FirstOrCreate(r)
~~~

Makefile
```diff
@@ -9,8 +9,8 @@ else
     b64_identity=$(shell echo '${identity}' | base64 -w 0 -)
 endif
 
-ros_ocp_msg='{"request_id": "uuid1234", "b64_identity": "test", "metadata": {"account": "2234", "org_id": "3340851", "source_id": "111", "cluster_uuid": "222", "cluster_alias": "name222"}, "files": ["http://localhost:8888/ros-ocp-usage.csv"]}'
```
upadhyeammit, line 12: Should we have both combinations of data? i.e with and without account?

Makefile
```diff
@@ -102,7 +102,7 @@ endif
 	sleep 5
 	$(eval SHAREURL=$(shell bin/mc share download --json myminio/insights-upload-perma/${CSVfile_name} | jq -r '.share'))
 	$(eval KAFKAPOD=$(shell oc get pods -o custom-columns=POD:.metadata.name --no-headers -n ${env} | grep kafka))
-	$(eval ros_ocp_msg_ephemeral = '{\"request_id\": \"uuid1234\", \"b64_identity\": \"test\", \"metadata\": {\"account\": \"2234\", \"org_id\": \"3340851\", \"source_id\": \"111\", \"cluster_uuid\": \"222\", \"cluster_alias\": \"name222\"}, \"files\": [\"$(SHAREURL)\"]}')
```
upadhyeammit, line 105: same, should we keep both?

internal/model/rh_account.go
patilsuraj767: Yes, That's correct

Makefile
patilsuraj767: I think it is better to have both as a part of unit test case, instead of having both in Makefile which we just use to ease the work in dev env.

Makefile
patilsuraj767: same as above comment

upadhyeammit (approved):

patilsuraj767:
Thanks for the review @upadhyeammit. Merging this PR.

PR #214, opened 2024-04-29
Fix partition deletion job

patilsuraj767:

Currently incorrect partition delete date(`partitionTableDate`) is getting set. 

- [x] Bugfix
- [ ] New Feature
- [ ] Refactor
- [ ] Unittests Added
- [ ] DRY code
- [ ] Dependency Added
- [ ] DB Migration Added

## Additional :mega:

### DIFF
--- a/internal/services/housekeeper/tablePartitionCleaner.go
+++ b/internal/services/housekeeper/tablePartitionCleaner.go
@@ -17,8 +17,8 @@ func DeletePartitions() {
 	retentionThresholdDate := currentTime.AddDate(0, 0, -cfg.DataRetentionPeriod)
 
 	// If the day of the month in $retentionThresholdDate is less than 15,
-	// set $partitionTableDate to the 16th of the previous month.
-	// Otherwise, set $partitionTableDate to the 1st of the current month.
+	// set $partitionTableDate to the 1st of the month.
+	// Otherwise, set $partitionTableDate to the 16th of the perivous month.
 	var partitionTableDate string
 	if retentionThresholdDate.Day() < 15 {
 		partitionTableDate = retentionThresholdDate.AddDate(0, 0, -retentionThresholdDate.Day()+1).Format("2006-01-02")
--- a/internal/services/housekeeper/tablePartitionCleaner.go
+++ b/internal/services/housekeeper/tablePartitionCleaner.go
@@ -21,9 +21,9 @@ func DeletePartitions() {
 	// Otherwise, set $partitionTableDate to the 1st of the current month.
 	var partitionTableDate string
 	if retentionThresholdDate.Day() < 15 {
-		partitionTableDate = time.Date(retentionThresholdDate.Year(), retentionThresholdDate.Month()-1, 16, 0, 0, 0, 0, retentionThresholdDate.Location()).Format("2006-01-02")
-	} else {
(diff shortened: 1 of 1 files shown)


internal/services/housekeeper/tablePartitionCleaner.go
```diff
@@ -17,13 +17,13 @@ func DeletePartitions() {
 	retentionThresholdDate := currentTime.AddDate(0, 0, -cfg.DataRetentionPeriod)
 
 	// If the day of the month in $retentionThresholdDate is less than 15,
-	// set $partitionTableDate to the 16th of the previous month.
-	// Otherwise, set $partitionTableDate to the 1st of the current month.
+	// set $partitionTableDate to the 1st of the month.
+	// Otherwise, set $partitionTableDate to the 16th of the perivous month.
```
r14chandra, line 21:
```suggestion
	// Otherwise, set $partitionTableDate to the 16th of the previous month.
```

r14chandra (changes requested):
Small typo

internal/services/housekeeper/tablePartitionCleaner.go
```diff
@@ -17,13 +17,13 @@ func DeletePartitions() {
...
 	var partitionTableDate string
 	if retentionThresholdDate.Day() < 15 {
-		partitionTableDate = time.Date(retentionThresholdDate.Year(), retentionThresholdDate.Month()-1, 16, 0, 0, 0, 0, retentionThresholdDate.Location()).Format("2006-01-02")
-	} else {
 		partitionTableDate = retentionThresholdDate.AddDate(0, 0, -retentionThresholdDate.Day()+1).Format("2006-01-02")
```
r14chandra, line 24: It should get the first day of the current month (as per the comment), right?
But for today's date, `partitionTableDate` is `2024-03-31` is it correct?

r14chandra: My bad! I mistakenly checked the previous code. Please ignore this comment!

patilsuraj767: So if we consider that job is going to run today then below should be the case.
currentTime = 2024-04-29
retentionThresholdDate = 2024-04-14
partitionTableDate = 2024-04-1 (so that all the partition before this date is removed)

internal/services/housekeeper/tablePartitionCleaner.go
```diff
@@ -17,13 +17,13 @@ func DeletePartitions() {
 	retentionThresholdDate := currentTime.AddDate(0, 0, -cfg.DataRetentionPeriod)
```
saltgen, line 17: Any thoughts on RDS time `us-east-1` or standard utc for currentTime instead of time.Now()?

saltgen (changes requested):

internal/services/housekeeper/tablePartitionCleaner.go
patilsuraj767: In the cost report we get timestamp in UTC, so make sense to convert this too in UTC.

saltgen (approved):

patilsuraj767:
Thanks @r14chandra and @saltgen for the review. Merging the PR.

patilsuraj767:
JIRA ticket - https://issues.redhat.com/browse/RHINENG-9786

PR #216, opened 2024-05-08
Fixes #RHINENG-9948 - Add CSV column datatype mapping

patilsuraj767:

golang sometime assert this value - "000004" as string and sometime as int. When it is assert as int then leading zeros are lot and we only get 4. 

- [x] Bugfix
- [ ] New Feature
- [ ] Refactor
- [ ] Unittests Added
- [ ] DRY code
- [ ] Dependency Added
- [ ] DB Migration Added

## Additional :mega:


internal/utils/aggregator.go
```diff
@@ -163,45 +164,11 @@ func elementsMatch(listA, listB interface{}) bool {
...
-		"memory_rss_usage_container_avg",
-		"memory_rss_usage_container_min",
-		"memory_rss_usage_container_max",
-		"memory_rss_usage_container_sum",
+	all_required_columns := make([]string, 0, len(types.CSVColumnMapping))
+	for k := range types.CSVColumnMapping {
+		all_required_columns = append(all_required_columns, k)
 	}
+
 	cloumns_in_csv := df.Names()
```
saltgen, line 172:
```suggestion
	columns_in_csv := df.Names()
```

internal/types/kruizePayload/updateResult.go
```diff
@@ -36,7 +37,10 @@ type ErrorData struct {
 
 func GetUpdateResultPayload(experiment_name string, containers []map[string]interface{}) []UpdateResult {
 	payload := []UpdateResult{}
-	df := dataframe.LoadMaps(containers)
+	df := dataframe.LoadMaps(
+		containers,
+		dataframe.WithTypes(types.CSVColumnMapping),
+	)
```
saltgen, line 43: Do we need this df creation in two places?

saltgen (changes requested):

internal/types/kruizePayload/updateResult.go
patilsuraj767: Yes, because `containers` value which it passed in this function is a map.

internal/types/csvColumnMapping.go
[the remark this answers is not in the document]
saltgen: I see your point, however let's avoid creating a file just for `csvColumnMapping`.
Since the processor need this, shall we store `csvColumnMapping` in `internal/services/common.go`?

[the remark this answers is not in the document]
patilsuraj767: We get the cyclic import issue if we shift `csvColumnMapping.go` to `types/kruizePayload/csvColumnMapping.go`, so for now let us keep it as  `types/csvColumnMapping.go`

saltgen (approved):
LGTM 👍🏼

saltgen:
@patilsuraj767 Shall I merge this up?

patilsuraj767:
Yes @saltgen

PR #217, opened 2024-05-15
RHINENG-9376 Assert workload before relating recommendation record

saltgen:
Looks like poller tries to save a recomm. for which workload was deleted potentially by the housekeeper service.
Additionally, I didn't find any workloads when searching with the ids from logs via Gabi.

- [ ] Yes
- [x] No

## Security Checklist :lock:

- [x] Bugfix
- [ ] New Feature
- [ ] Refactor
- [ ] Unittests Added
- [x] DRY code
- [ ] Dependency Added
- [ ] DB Migration Added

## Additional :mega:

More details added in JIRA card

### DIFF
--- a/internal/services/recommendation_poller.go
+++ b/internal/services/recommendation_poller.go
@@ -182,6 +182,8 @@ func PollForRecommendations(msg *kafka.Message, consumer_object *kafka.Consumer)
 			}
 			return
 		}
+	} else {
+		commitKafkaMsg(msg, consumer_object)
 	}
 
 }
--- a/internal/services/recommendation_poller.go
+++ b/internal/services/recommendation_poller.go
@@ -151,29 +151,34 @@ func PollForRecommendations(msg *kafka.Message, consumer_object *kafka.Consumer)
 		log.Errorf("Error while checking for recommendation_set record: %s", err)
 		return
 	}
-	recommendationFound := !reflect.ValueOf(recommendation_stored_in_db).IsZero()
+	workloadExists := model.WorkloadExistsByID(workloadID)
 
-	switch recommendationFound {
-	case false:
-		poll_cycle_complete := requestAndSaveRecommendation(kafkaMsg, "New")
-		if poll_cycle_complete {
-			commitKafkaMsg(msg, consumer_object)
-		}
-		// To consume upcoming Kafka msg, explicitly
-		return
-	case true:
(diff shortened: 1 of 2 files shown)


patilsuraj767 (approved):
LGTM 👍

internal/services/recommendation_poller.go
```diff
@@ -151,29 +151,34 @@ func PollForRecommendations(msg *kafka.Message, consumer_object *kafka.Consumer)
...
+				} else {
 					commitKafkaMsg(msg, consumer_object)
 				}
-			} else {
-				commitKafkaMsg(msg, consumer_object)
 			}
+			return
 		}
-		return
 	}
```
patilsuraj767, line 182: Shouldn't we have else block here and commit the kafka message? because autocommit is set to false in poller service.

patilsuraj767 (changes requested):

internal/services/recommendation_poller.go
saltgen: If MonitoringEndTime.UTC() defaults to 0001-01-01 00:00:00 +0000 UTC, then I would assume there is some issue with Kruize, in that case I don't think we should commit the msg.
However I am happy to add a warning log before `return`, does that sound good?

saltgen: I will add a warn log as well in the else, thanks!

saltgen: Change pushed

internal/services/recommendation_poller.go
```diff
@@ -151,29 +151,37 @@ func PollForRecommendations(msg *kafka.Message, consumer_object *kafka.Consumer)
...
 				}
 			} else {
 				commitKafkaMsg(msg, consumer_object)
+				log.Warn("monitoring_end_time is set to 0001-01-01 00:00:00 +0000; recommendationID: ", recommendation_stored_in_db.ID)
 			}
+			return
 		}
-		return
 	}
+
```
patilsuraj767, line 186: We need `else` block here on line 186. So if `workloadExists == false` we should commit the current kafka message.

patilsuraj767 (changes requested):

patilsuraj767 (approved):
Not tested locally but overall Looks good to me 👍

yash2189:
Tested the changes on EE and did not observe any errors mentioned in description of the bug card. Added my observations

PR #219, opened 2024-05-31
RHINENG-10255 Unit conversion on ROS OCP backend

saltgen:

- Reverted memory unit from Mi to bytes of container usage data to Kruize
- Box plots data is required on recommendation by ID endpoint, removed from list endpoint
- Unit conversion now supported on recommendation ID endpoint through optional params

- [ ] Yes
- [x] No

## Security Checklist :lock:

- [ ] Bugfix
- [x] New Feature
- [x] Refactor
- [ ] Unittests Added
- [ ] DRY code
- [ ] Dependency Added
- [ ] DB Migration Added

## Additional :mega:

Summary on latency comparison available in JIRA.

### DIFF
--- a/internal/api/handlers.go
+++ b/internal/api/handlers.go
@@ -16,6 +16,12 @@ func GetRecommendationSetList(c echo.Context) error {
 	XRHID := c.Get("Identity").(identity.XRHID)
 	OrgID := XRHID.Identity.OrgID
 	user_permissions := get_user_permissions(c)
+	handlerName := "recommendationset-list"
+	unitChoices := map[string]string{
+		"cpu":    "cores",
+		"memory": "bytes",
+	}
+
 	var orderHow string
 	var orderBy string
 	// Default values
@@ -96,7 +102,7 @@ func GetRecommendationSetList(c echo.Context) error {
 		recommendationData["workload"] = recommendation.Workload.WorkloadName
 		recommendationData["container"] = recommendation.ContainerName
 		recommendationData["last_reported"] = recommendation.Workload.Cluster.LastReportedAtStr
-		recommendationData["recommendations"] = UpdateRecommendationJSON(recommendation.ID, recommendation.Workload.Cluster.ClusterUUID, recommendation.Recommendations)
+		recommendationData["recommendations"] = UpdateRecommendationJSON(handlerName, recommendation.ID, recommendation.Workload.Cluster.ClusterUUID, unitChoices, recommendation.Recommendations)
 		allRecommendations = append(allRecommendations, recommendationData)
 
 	}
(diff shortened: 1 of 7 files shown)


internal/api/utils.go
```diff
@@ -404,7 +455,25 @@ func filterNotifications(recommendationID string, clusterUUID string, recommenda
...
+
+	for _, period := range []string{"short_term", "medium_term", "long_term"} {
+		intervalData, ok := recommendation_terms[period].(map[string]interface{})
+		if !ok {
+			continue
+		}
+		delete(intervalData, "plots")
+	}
+	return recommendationJSON
+}
```
upadhyeammit, line 474: We can leverage Postgresql capabilities here, it took me sometime however there is way to exclude the plots data in select statement itself,
~~~
# select recommendations #- '{recommendation_terms,long_term,plots}' #- '{recommendation_terms,medium_term,plots}' #- '{recommendation_terms,short_term,plots}' from recommendation_sets;
~~~

On the top of this we can create a view so recommendation-set only queries from that view,
~~~
# create view recommendation_sets_no_box_plots as select id, workload_id, container_name, monitoring_start_time, monitoring_end_time, recommendations #- '{recommendation_terms,long_term,plots}' #- '{recommendation_terms,medium_term,plots}' #- '{recommendation_terms,short_term,plots}' as recommendations from recommendation_sets;
~~~

Locally I tested this with notifications as its present in all three terms and it works,
~~~
# select recommendations #- '{recommendation_terms,long_term,notifications}' #- '{recommendation_terms,medium_term,notifications}' #- '{recommendation_terms,short_term,notifications}' from recommendation_sets;
~~~

openapi.json
```diff
@@ -171,6 +175,35 @@
               "type": "string"
             },
             "description": "The recommendation UUID"
+          },
+          {
```
kgaikwad, line 179: @saltgen  - could you check here - https://swagger.io/docs/specification/describing-parameters/#default and see if it helps to present default value for these params?

patilsuraj767 (approved):
Not reviewed `openapi.json` but rest all changes looks good to me and also tested locally it is working as expected. 👍  
Thanks @saltgen

openapi.json
saltgen: Yep, possible via the enum type, changes pushed

internal/api/utils.go
saltgen: As discussed, let's take this up in a separate iteration.

internal/api/handlers.go
```diff
@@ -114,13 +120,49 @@ func GetRecommendationSet(c echo.Context) error {
...
+
+	if memoryUnitParam != "" {
+		if !memoryUnitOptions[memoryUnitParam] {
+			return c.JSON(http.StatusBadRequest, echo.Map{"status": "error", "message": "invalid memory unit"})
+		} else {
+			unitChoices["memory"] = memoryUnitParam
+		}
+	} else {
+		unitChoices["memory"] = "MiB"
+	}
```
upadhyeammit, line 164: Not a blocker, maybe we can create a common func for memory and cpu.

internal/api/utils.go
```diff
@@ -217,24 +210,47 @@ func truncateComponentUnits(recommendationJSON map[string]interface{}) map[strin
...
-			cpu["amount"] = truncateToThreeDecimalPlaces(cpuInCores)
-			cpu["format"] = nil
-		}
-		return ok
+	if cpuUnit == "millicores" {
+		convertedValueCPU = math.Round(cpuValue * 1000) // millicore values don't require decimal precision
+	} else if cpuUnit == "cores" {
+		convertedValueCPU = truncateToThreeDecimalPlaces(cpuValue)
+	} else {
+		convertedValueCPU = cpuValue
```
upadhyeammit, line 220: Nit and not a blocker, considering our default we could have checked for cpuUnit == "cores" first, just to save one comparison

internal/api/utils.go
```diff
@@ -308,6 +326,36 @@ func truncateComponentUnits(recommendationJSON map[string]interface{}) map[strin
...
+						if memoryUsage, ok := datapointMap["memoryUsage"].(map[string]interface{}); ok {
+							memoryUnit := unitsToTransform["memory"]
+							if _, ok := memoryUsage["format"].(string); ok {
+								memoryUsage["format"] = memoryUnit
+							}
+							for _, key := range []string{"q1", "q3", "min", "max", "median"} {
+								memoryValue, _ := memoryUsage[key].(float64)
+								memoryUsage[key] = convertMemoryUnit(memoryUnit, memoryValue)
+							}
+						}
```
upadhyeammit, line 352: Future scope, not a blocker, maybe if there is performance challenge we can run cpu and memory conversations as go routine.

upadhyeammit (approved):
It looks good to me, I have just added some suggestions and those are not blockers.

internal/api/handlers.go
saltgen: each does things a bit differently from another, for instance cpu values require truncation whereas memory doesn't, hence the separation

internal/api/utils.go
saltgen: Sure, we could look into it

internal/api/handlers.go
upadhyeammit: Apologies, I should have highlighted the block! I mean below blocks look identical,
~~~
	if cpuUnitParam != "" {
		if !cpuUnitOptions[cpuUnitParam] {
			return c.JSON(http.StatusBadRequest, echo.Map{"status": "error", "message": "invalid cpu unit"})
		} else {
			unitChoices["cpu"] = cpuUnitParam
		}
	} else {
		unitChoices["cpu"] = "cores"
	}

	if memoryUnitParam != "" {
		if !memoryUnitOptions[memoryUnitParam] {
			return c.JSON(http.StatusBadRequest, echo.Map{"status": "error", "message": "invalid memory unit"})
		} else {
			unitChoices["memory"] = memoryUnitParam
		}
	} else {
		unitChoices["memory"] = "MiB"
	}
~~~

PR #223, opened 2024-06-25
Fixes #RHCLOUD-33105 - Modified kafka consumer config for non-SSL broker

patilsuraj767:
As per comment on the JIRA ticket we need to add custom kafka consumer config for non-auth environments. 

- [ ] Bugfix
- [ ] New Feature
- [x] Refactor
- [ ] Unittests Added
- [ ] DRY code
- [ ] Dependency Added
- [ ] DB Migration Added

## Additional :mega:

### DIFF
--- a/internal/kafka/consumer.go
+++ b/internal/kafka/consumer.go
@@ -39,8 +39,6 @@ func StartConsumer(kafka_topic string, handler func(msg *kafka.Message, consumer
 			"enable.auto.commit":       auto_commit,
 			"go.logs.channel.enable":   true,
 			"allow.auto.create.topics": true,
-			"session.timeout.ms":       120000,
-			"heartbeat.interval.ms":    30000,
 		}
 
 		// As per librdkafka doc - https://github.com/confluentinc/librdkafka/blob/master/CONFIGURATION.md?plain=1#L73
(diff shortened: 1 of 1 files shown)


internal/kafka/consumer.go
```diff
@@ -59,6 +57,9 @@ func StartConsumer(kafka_topic string, handler func(msg *kafka.Message, consumer
 		}
 	}
 
+	configMap["session.timeout.ms"] = 120000
+	configMap["heartbeat.interval.ms"] = 30000
```
saltgen, line 61: Could we move these inside the configMap var on L#51?

saltgen (changes requested):

internal/kafka/consumer.go
patilsuraj767: No, we can't because when PROD will move to AWS Managed Kafka(migration on 27th June) then these config will not get applied if we move these changes to L#51.

upadhyeammit (approved):

internal/kafka/consumer.go
saltgen: Ah the else clause, my bad

saltgen (approved):

PR #224, opened 2024-07-01
RHINENG-10671  API should return k8s units for values

saltgen:
Update 2:

We are switching to a boolean value param, i.e. `true-units=true/false`.
This is due to an OpenAPI limitation, there is no native way to define a value-less boolean param in the spec.

--------------------------------------------------

Update:

Since we are switching to k8s units by default, `k8s-units` has been replaced by `true-units`.
It is a boolean param, presence denotes `true`. Since we can expect a k8s service to show values in k8s units the introduction and usage of `true-units` makes a bit more sense.
Thank you @kgaikwad for the suggestion.

---------------------------------------------------
> The API is reporting values in K8S supported units for the UI to avoid any unit string manipulation in the recommendation section.
> 
> Now that we have box plots data as well, this approach no longer seems to work as expected.

As per internal discussions, the API will now return k8s units by default.

This PR includes another query parameter `k8s-units` ("true" by default).
In case users require values in real-world units the same can be set to "false" while sending the request.

- [ ] Yes
- [x] No

## Security Checklist :lock:

- [ ] Bugfix
- [ ] New Feature
- [x] Refactor
- [ ] Unittests Added
- [ ] DRY code
- [ ] Dependency Added
- [ ] DB Migration Added

## Additional :mega:

Please check pinned comment on JIRA card for more details.

### DIFF
--- a/internal/api/handlers.go
+++ b/internal/api/handlers.go
@@ -118,6 +118,18 @@ func GetRecommendationSetList(c echo.Context) error {
 		log.Error("unable to fetch records from database", error)
 	}
 
+	trueUnitsStr := c.QueryParam("true-units")
+	var trueUnits bool
+
+	if trueUnitsStr != "" {
+		trueUnits, err = strconv.ParseBool(trueUnitsStr)
+		if err != nil {
+			return c.JSON(http.StatusBadRequest, echo.Map{"status": "error", "message": "invalid value for true-units"})
+		}
+	}
+	
+	setk8sUnits := !trueUnits
+
 	allRecommendations := []map[string]interface{}{}
 
 	for _, recommendation := range recommendationSets {
@@ -132,7 +144,7 @@ func GetRecommendationSetList(c echo.Context) error {
 		recommendationData["workload"] = recommendation.Workload.WorkloadName
 		recommendationData["container"] = recommendation.ContainerName
 		recommendationData["last_reported"] = recommendation.Workload.Cluster.LastReportedAtStr
-		recommendationData["recommendations"] = UpdateRecommendationJSON(handlerName, recommendation.ID, recommendation.Workload.Cluster.ClusterUUID, unitChoices, recommendation.Recommendations)
+		recommendationData["recommendations"] = UpdateRecommendationJSON(handlerName, recommendation.ID, recommendation.Workload.Cluster.ClusterUUID, unitChoices, setk8sUnits, recommendation.Recommendations)
 		allRecommendations = append(allRecommendations, recommendationData)
 
 	}
@@ -159,6 +171,18 @@ func GetRecommendationSet(c echo.Context) error {
 	}
 
 	var unitChoices = make(map[string]string)
+
+	trueUnitsStr := c.QueryParam("true-units")
+	var trueUnits bool
+
+	if trueUnitsStr != "" {
+		trueUnits, err = strconv.ParseBool(trueUnitsStr)
+		if err != nil {
+			return c.JSON(http.StatusBadRequest, echo.Map{"status": "error", "message": "invalid value for true-units"})
+		}
+	}
+	
+	setk8sUnits := !trueUnits
 
 	cpuUnitParam := c.QueryParam("cpu-unit")
 	var cpuUnitOptions = map[string]bool{
@@ -212,7 +236,7 @@ func GetRecommendationSet(c echo.Context) error {
 		recommendationSlice["workload"] = recommendationSet.Workload.WorkloadName
 		recommendationSlice["container"] = recommendationSet.ContainerName
 		recommendationSlice["last_reported"] = recommendationSet.Workload.Cluster.LastReportedAtStr
-		recommendationSlice["recommendations"] = UpdateRecommendationJSON(handlerName, recommendationSet.ID, recommendationSet.Workload.Cluster.ClusterUUID, unitChoices, recommendationSet.Recommendations)
+		recommendationSlice["recommendations"] = UpdateRecommendationJSON(handlerName, recommendationSet.ID, recommendationSet.Workload.Cluster.ClusterUUID, unitChoices, setk8sUnits, recommendationSet.Recommendations)
 	}
 
 	return c.JSON(http.StatusOK, recommendationSlice)
--- a/openapi.json
+++ b/openapi.json
@@ -69,6 +69,15 @@
             }
           },
           {
+            "name": "true-units",
+            "in": "query",
+            "required": false,
+            "schema": {
+              "type": "boolean"
+            },
+            "description": "Shows all values in true/real-world units. Accepts 1, t, T, TRUE, true, True, 0, f, F, FALSE, false, False."
+          },
+          {
             "name": "start_date",
             "in": "query",
             "description": "Start date",
@@ -204,6 +213,15 @@
               "type": "string"
             },
             "description": "The recommendation UUID"
+          },
+          {
+            "name": "true-units",
+            "in": "query",
+            "required": false,
+            "schema": {
+              "type": "boolean"
+            },
+            "description": "Shows all values in true/real-world units. Accepts 1, t, T, TRUE, true, True, 0, f, F, FALSE, false, False."
           },
           {
             "in": "query",
--- a/internal/api/utils.go
+++ b/internal/api/utils.go
@@ -242,7 +242,7 @@ func convertMemoryUnit(memoryUnit string, memoryValue float64) float64 {
 
 }
 
-func transformComponentUnits(unitsToTransform map[string]string, recommendationJSON map[string]interface{}) map[string]interface{} {
+func transformComponentUnits(unitsToTransform map[string]string, updateUnitsk8s bool, recommendationJSON map[string]interface{}) map[string]interface{} {
 	/*
 		Truncates CPU units(cores) to three decimal places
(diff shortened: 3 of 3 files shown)


upadhyeammit (approved):

kgaikwad:
@saltgen, as I can see it is approved already, what is blocking this PR to get merge?

saltgen:
@kgaikwad I had to drop a noti. for FE before merging this PR. Then again just realized I need to push the openapi changes here.

internal/api/handlers.go
```diff
@@ -130,6 +141,17 @@ func GetRecommendationSet(c echo.Context) error {
...
+	k8sUnits := c.QueryParam("k8s-units")
+	var setk8sUnits bool
+
+	if k8sUnits == "" || k8sUnits == "true" {
+		setk8sUnits = true
+	} else if k8sUnits == "false" {
+		setk8sUnits = false
+	} else {
+		return c.JSON(http.StatusBadRequest, echo.Map{"status": "error", "message": "invalid option for k8s-units"})
+	}
```
upadhyeammit, line 153: This can be moved to function as its getting repeated twice?

saltgen: I don't quite see the advantage, since we don't have any other endpoints

upadhyeammit: okay, maybe I would let @patilsuraj767 and @kgaikwad add their thoughts as well?
more about DRY: https://www.geeksforgeeks.org/dont-repeat-yourselfdry-in-software-development/ there are multiple refs about it, however I felt this post is bit clear

saltgen: We'll also need to modularize [this](https://github.com/RedHatInsights/ros-ocp-backend/blob/main/internal/api/handlers.go#L161-L193) in that case, let's plan it for later when we start work on the plots endpoint.

kgaikwad: > maybe I would let @patilsuraj767 and @kgaikwad add their thoughts as well?

+1 from my side for extracting duplicate code blocks to function.

@saltgen - Yes, there are definitely changes required to support individual usage(CPU/Mem) units based on `k8sUnits` param selection which are currently missing here. But at least I don't understand why do you want to wait for ^^ to extract out this code block.

upadhyeammit:
Hello @saltgen can you rebase the changes to main, to actually include https://github.com/RedHatInsights/ros-ocp-backend/pull/255 . I like to see how it plays with this change! I can see that you can conditionally handle the situations when both parameters are available, however for assurance like to see. Thanks!

internal/api/handlers.go
saltgen: @kgaikwad I'm not in favor of wrapping almost all logic in functions because function calls have a performance overhead. I believe we should try to achieve a balance.

1. I think we are taking DRY too literally here. We have exactly two usages of the this block of code, because there are two endpoints.

2. We might have the box plots data endpoint soon, then we could plan this block and individual usage(CPU/Mem) units in one go and make something like [this](https://github.com/RedHatInsights/ros-ocp-backend/blob/main/internal/api/utils.go#L67).

Let me know what I can do to unblock this PR.

openapi.json
```diff
@@ -205,6 +219,20 @@
...
+            "required": false,
+            "schema": {
+              "type": "string",
+              "enum": [
+                "true",
+                "false"
+              ],
+              "default": "true"
+            }
+          },
```
kgaikwad, line 235:
```suggestion
          {
              "name": "k8s-units",
              "in": "query",
              "description": "If true, the endpoint returns values in k8s-units formats",
              "example": false,
              "required": false,
              "allowEmptyValue": true,
              "schema": {
                    "type": "boolean",
                    "default": true
                }
          },
```
Why is it a `string` type parameter?

internal/api/handlers.go
kgaikwad: > I'm not in favor of wrapping almost all logic in functions because function calls have a performance overhead.

@saltgen,  When I see this function `GetRecommendationSetList` - https://github.com/RedHatInsights/ros-ocp-backend/blob/1920484c0f5de14bf1aaf8dd094ed679b82035ac/internal/api/handlers.go#L15-L158
Not each logic wrapped into multiple functions and suggestion provided by @upadhyeammit above didn't talk about wrapping each & every logic inside functions.
My vote to the suggestion is purely for code readability and maintenance perspective. In fact, I see repetitive logic at multiple places in this same file which can be extracted into a single function. I want to understand your concern here - Could you please elaborate what kind of performance overhead you are anticipating here with this suggestion? Will it be possible to pass on few references for my understanding?

> We might have the box plots data endpoint soon

Could you provide reference ticket to get when it is planned for and how these both things are correlated?

kgaikwad (dismissed):

openapi.json
saltgen: Values passed to query parameters are converted to `string` by default, from the client level.

internal/api/handlers.go
saltgen: Moving conversation to internal DM

openapi.json
saltgen: Native boolean support can be done if we are accepting a json payload as such.
```
{
    "isActive": true
}
```
In case of query parameters, the values will be string when captured on the backend.

saltgen: @kgaikwad Since the API spec is user-facing, if we denote the type as boolean then the user might get confused about the following,

1. Boolean in which language, JSON/Go/Python etc.
2. How to specify the type as boolean on the query paramm?

kgaikwad: OK! In this case, will it make sense to check presence of `k8-units` query param as flag? With this there will be only two possibilities to verify - `k8-units` exist or not exist.

saltgen: This is a possibility, but then UI will need to provide the `k8s-units` parameter on the endpoint.
Otherwise the param value would not be truthy, so we won't be able to provide k8s units as default.

upadhyeammit:
Few observations after the rebase,
1. If we keep k8s-units as true or dont provide any value to it(default) then we still accept memory-unit and cpu-unit params with non k8s-units values(MiB/GiB). For memory-unit I can see GiB, MiB is accepted however the output is in Mi/GI. So I think as user its not correct experience? For this I think the memory-unit and cpu-unit should accept values as per parent param k8s-units?

2. This does also mean we need to accept memory-unit and cpu-unit compliant to k8s-units? 

3. The invalid errors should also show what are the valid values for the params? Instead of just saying its invalid, so if someone is providing not matching units as per k8s-units param then there is enough detail about where is the problem?
~~~
{
    "message": "invalid memory unit",
    "status": "error"
}
~~~

openapi.json
kgaikwad: As we want to keep k8-units as default, it makes sense to add parameter to override default behavior.  In this way, UI will not be require to include it in every request.  You can rename something as long as it conveys meaning of & intention behind this parameter to end users.

saltgen: Yep, I have pushed the changes and updated PR description, new param is `true-units`.

saltgen:
> Few observations after the rebase,
> 
> 1. If we keep k8s-units as true or dont provide any value to it(default) then we still accept memory-unit and cpu-unit params with non k8s-units values(MiB/GiB). For memory-unit I can see GiB, MiB is accepted however the output is in Mi/GI. So I think as user its not correct experience? For this I think the memory-unit and cpu-unit should accept values as per parent param k8s-units?
> 2. This does also mean we need to accept memory-unit and cpu-unit compliant to k8s-units?
> 3. The invalid errors should also show what are the valid values for the params? Instead of just saying its invalid, so if someone is providing not matching units as per k8s-units param then there is enough detail about where is the problem?
> 
> ```
> {
>     "message": "invalid memory unit",
>     "status": "error"
> }
> ```

1. Since the implementation has now changed from `k8s-units` to `true-units`, this shouldn't be a problem anymore, basically when `true-units` is present we internally set `setk8sUnits = false`.
2. `memory-unit` and `cpu-unit` values will now always be in k8s units unless `true-units` is present as a query param.
3. I believe the user ought to check the available options from the OpenAPI spec.

Additionally, I think this simplifies the whole approach. 
If the user provides the `true-units` param, all values should be expected in real-world units else k8s-units as default.

saltgen:
> Few observations after the rebase,
> 
> 1. If we keep k8s-units as true or dont provide any value to it(default) then we still accept memory-unit and cpu-unit params with non k8s-units values(MiB/GiB). For memory-unit I can see GiB, MiB is accepted however the output is in Mi/GI. So I think as user its not correct experience? For this I think the memory-unit and cpu-unit should accept values as per parent param k8s-units?
> 2. This does also mean we need to accept memory-unit and cpu-unit compliant to k8s-units?
> 3. The invalid errors should also show what are the valid values for the params? Instead of just saying its invalid, so if someone is providing not matching units as per k8s-units param then there is enough detail about where is the problem?
> 
> ```
> {
>     "message": "invalid memory unit",
>     "status": "error"
> }
> ```

Additionally, regarding point 1, the users might provide a value as MiB for memory-unit, but as per the implmentation k8s unit equivalent is expected for the same.

saltgen:
The OpenAPI changes need re-work, the specification does not natively support valueless query params.
https://github.com/OAI/OpenAPI-Specification/issues/1782

saltgen:
@kgaikwad @upadhyeammit All changes pushed

PR #236, opened 2024-07-15
Fixes #RHINENG-10738 - Batch insert workload metrics

patilsuraj767:

This is required to reduce the db commits/transactions.

- [ ] Bugfix
- [ ] New Feature
- [x] Refactor
- [ ] Unittests Added
- [ ] DRY code
- [ ] Dependency Added
- [ ] DB Migration Added

## Additional :mega:

### DIFF
--- a/internal/services/report_processor.go
+++ b/internal/services/report_processor.go
@@ -145,6 +145,7 @@ func ProcessReport(msg *kafka.Message, _ *kafka.Consumer) {
 					continue
 				}
 
+				workload_metric_arr := []model.WorkloadMetrics{}
 				for _, data := range usage_data_byte {
 
 					interval_start_time, err := utils.ConvertISO8601StringToTime(data.Interval_start_time)
@@ -172,12 +173,12 @@ func ProcessReport(msg *kafka.Message, _ *kafka.Consumer) {
 							IntervalEnd:   interval_end_time,
 							UsageMetrics:  container_usage_metrics,
 						}
-						if err := workload_metric.CreateWorkloadMetrics(); err != nil {
-							log.Errorf("unable to add record to workload_metrics table: %v. Error: %v", workload_metric, err)
-							continue
-						}
+						workload_metric_arr = append(workload_metric_arr, workload_metric)
 					}
 
+				}
+				if err := model.BatchInsertWorkloadMetrics(workload_metric_arr, rh_account.OrgId); err != nil {
+					log.Errorf("unable to batch insert to workload_metrics table. Error: %v", err)
 				}
 			}
 			// Sending recommendation request to recommendation-poller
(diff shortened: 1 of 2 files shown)


internal/services/report_processor.go
```diff
@@ -172,13 +173,13 @@ func ProcessReport(msg *kafka.Message, _ *kafka.Consumer) {
 							IntervalEnd:   interval_end_time,
 							UsageMetrics:  container_usage_metrics,
 						}
-						if err := workload_metric.CreateWorkloadMetrics(); err != nil {
-							log.Errorf("unable to add record to workload_metrics table: %v. Error: %v", workload_metric, err)
-							continue
-						}
+						workload_metric_arr = append(workload_metric_arr, workload_metric)
```
upadhyeammit, line 176: When we are unable to parse `interval_start_time` or `interval_end_time` then we skip the iteration and go to next one. However for `container_usage_metrics` I see we dont have `continue`,
~~~
container_usage_metrics, err := json.Marshal(container.Metrics)
if err != nil {
  log.Errorf("Unable to marshal container usage data: %v", err)
}
~~~

Considering the fact that nothing in `workload_metrics` is nullable, we should also use continue in above if block? so this way we don't error out while inserting or updating in database later in BatchInsertWorkloadMetrics? and if I understand correctly if one of the row is missing `workload_metrics` then it won't process whole batch?

P.S its not related to this pull request, however as we are switching to array I think this can affect data insertion or update if there are some rows which are missing the fields

patilsuraj767: I agree. I will add `continue` for `container_usage_metrics` block.

patilsuraj767: Done

internal/services/report_processor.go
```diff
@@ -172,13 +174,13 @@ func ProcessReport(msg *kafka.Message, _ *kafka.Consumer) {
...
 					}
 
 				}
+				if err := model.BatchInsertWorkloadMetrics(workload_metric_arr, rh_account.OrgId); err != nil {
+					log.Errorf("unable to batch insert to workload_metrics table. Error: %v", err)
```
saltgen, line 182: @patilsuraj767 Any thoughts on completing the transaction here i.e. after we know there was an error with BatchInsertWorkloadMetrics?

patilsuraj767: Not exactly sure what you mean by completing transaction here. DB transaction takes place inside the model i.e. inside `BatchInsertWorkloadMetrics` function.

saltgen: I see, I am suggesting something like below, in case there is an db insert error, where ever we are writing to DB.
https://github.com/RedHatInsights/ros-ocp-backend/blob/main/internal/services/recommendation_poller.go#L44

saltgen: This might require some more code, but we will be better off in the long run since bulk insert might prove to be expensive.

patilsuraj767: As discussed internally in the meeting above suggestion are good to implement but as in future we will be moving to Thanos for storing the workload metrics and currently ROSOCP itself does not use workload_metrics data. So we are not going ahead and implementing the db transaction logic.

yash2189:
LGTM from QE side testing on EE

PR #243, opened 2024-07-30
RHINENG-11430 Metrics for HTTP request-response cycle

saltgen:

Capture the following metrics w.r.t request and response cycle between ROS OCP and Kruize on Prometheus.

- kruize_create_experiment_request_total
- kruize_update_result_request_total
- rosocp_recommendation_request_total
- rosocp_recommendation_success_total

- [ ] Yes
- [x] No

## Security Checklist :lock:

- [ ] Bugfix
- [x] New Feature
- [ ] Refactor
- [ ] Unittests Added
- [ ] DRY code
- [ ] Dependency Added
- [ ] DB Migration Added

## Additional :mega:
image

Please check JIRA card for additional details.

### DIFF
--- a/internal/services/recommendation_poller.go
+++ b/internal/services/recommendation_poller.go
@@ -54,6 +54,7 @@ func transactionForRecommendation(recommendationSetList []model.RecommendationSe
 			return err
 		}
 	}
+	recommendationSuccess.Inc()
 	return tx.Commit().Error
 }
 
@@ -72,6 +73,7 @@ func requestAndSaveRecommendation(kafkaMsg types.RecommendationKafkaMsg, recomme
 		log.Errorf("unable to list recommendation for: %v", err)
 		return poll_cycle_complete
 	}
+	recommendationRequest.Inc()
 
 	containers := recommendation[0].Kubernetes_objects[0].Containers
 	recommendationSetList := []model.RecommendationSet{}
--- a/internal/services/recommendation_poller.go
+++ b/internal/services/recommendation_poller.go
@@ -54,7 +54,6 @@ func transactionForRecommendation(recommendationSetList []model.RecommendationSe
 			return err
 		}
 	}
-	recommendationSuccess.Inc()
 	return tx.Commit().Error
 }
 
@@ -116,6 +115,7 @@ func requestAndSaveRecommendation(kafkaMsg types.RecommendationKafkaMsg, recomme
 		txError := transactionForRecommendation(recommendationSetList, histRecommendationSetList, experiment_name, recommendationType)
 		if txError == nil {
 			poll_cycle_complete = true
+			recommendationSuccess.Inc()
 		} else {
 			poll_cycle_complete = false
 		}
--- a/internal/services/recommendation_poller.go
+++ b/internal/services/recommendation_poller.go
@@ -68,6 +68,7 @@ func requestAndSaveRecommendation(kafkaMsg types.RecommendationKafkaMsg, recomme
 		end_interval := utils.ConvertDateToISO8601(maxEndTimeFromReport.String())
 		if err.Error() == fmt.Sprintf("Recommendation for timestamp - \" %s \" does not exist", end_interval) {
 			log.Infof("recommendation does not exist for timestamp - \" %s \"", end_interval)
+			recommendationRequest.Inc()
 		}
 		log.Errorf("unable to list recommendation for: %v", err)
 		return poll_cycle_complete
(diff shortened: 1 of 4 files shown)


internal/utils/kruize/metrics.go
```diff
@@ -16,4 +16,12 @@ var (
 		Name: "rosocp_invalid_recommendation_total",
 		Help: "The total number of invalid recommendation send by Kruize",
 	})
+	createExperimentRequest = promauto.NewCounter(prometheus.CounterOpts{
+		Name: "kruize_create_experiment_request_total",
+		Help: "The total number of experiments created on Kruize",
```
patilsuraj767, line 21:
```suggestion
		Help: "The total number of createExperiment request sent to Kruize",
```

internal/utils/kruize/kruize_api.go
```diff
@@ -84,7 +84,7 @@ func Create_kruize_experiments(experiment_name string, cluster_identifier string
 			return nil, fmt.Errorf("%s", resdata["message"])
 		}
 	}
-
+	createExperimentRequest.Inc()
```
patilsuraj767, line 87: This will only increment if createExperiment request was successfully. To track all the request `createExperimentRequest.Inc()` should be on line 57.

internal/utils/kruize/kruize_api.go
```diff
@@ -139,6 +139,7 @@ func Update_results(experiment_name string, payload_data []kruizePayload.UpdateR
 		}
 	}
 
+	updateResultRequest.Inc()
```
patilsuraj767, line 142: same as above. It should be on line 110

internal/services/recommendation_poller.go
```diff
@@ -72,6 +73,7 @@ func requestAndSaveRecommendation(kafkaMsg types.RecommendationKafkaMsg, recomme
 		log.Errorf("unable to list recommendation for: %v", err)
 		return poll_cycle_complete
 	}
+	recommendationRequest.Inc()
```
patilsuraj767, line 76: I think this should be inside the function definition of `Update_recommendations` there can be many reason that ROSOCP send recommendationRequest to kruize but code never reaches here to increment the counter. One example is if ROSOCP is unable to unmarshal the API response.

internal/services/recommendation_poller.go
```diff
@@ -54,6 +54,7 @@ func transactionForRecommendation(recommendationSetList []model.RecommendationSe
 			return err
 		}
 	}
+	recommendationSuccess.Inc()
```
patilsuraj767, line 57: As the `recommendationSuccess` metric says it is `The total number of recommendations saved by ROSOCP`, This increment should happen after committing the transaction.

internal/utils/kruize/metrics.go
```diff
@@ -16,4 +16,12 @@ var (
...
+	})
+	updateResultRequest = promauto.NewCounter(prometheus.CounterOpts{
+		Name: "kruize_update_result_request_total",
+		Help: "The total number of requests/csvs sent to Kruize for UpdateResult",
```
patilsuraj767, line 25:
```suggestion
		Help: "The total number of requests sent to Kruize for UpdateResult",
```
We do not sent whole CSV in a single request, can be bit misleading.

patilsuraj767 (changes requested):

internal/services/recommendation_poller.go
saltgen: I think we should update this counter post usage of `Update_recommendations`, if in case there are errors encountered, we can explicitly handle the same in the poller service/where ever we used the function.

patilsuraj767: But then we won't be getting the exact count i.e. total number of times we hit the update_recommendation API. If you see the error block above which says `recommendation does not exist for timestamp` I think in this case as well we should increment the prometheus counter.

saltgen: Change pushed

saltgen: Change pushed

patilsuraj767 (approved):
LGTM 👍 Thanks @saltgen

saltgen:
@yash2189 Please take a look at the PR check failures, we don't require testing on these changes.

yash2189:
I can see the recommendations were not received in time hence tests were failing

yash2189:
If you want I can try to deploy on EE and manually run one of the tests

saltgen:
I don't think that would be necessary, thanks.

PR #283, opened 2024-12-09
Fixes #RHINENG-14713 - Error if no valid workload_type found

upadhyeammit:
## PR Title :boom:

Suggested formats: 

2. RHIROS-XXX Title
3. RHINENG-XXX Title 

Please include the __context of this change__ here. If the change depends on Kruize add details about that as well!

- [ ] Yes
- [ ] No

## Security Checklist :lock:

- [ ] Bugfix
- [ ] New Feature
- [ ] Refactor
- [ ] Unittests Added
- [ ] DRY code
- [ ] Dependency Added
- [ ] DB Migration Added

## Additional :mega:

### DIFF
--- a/internal/utils/aggregator.go
+++ b/internal/utils/aggregator.go
@@ -31,11 +31,7 @@ func Aggregate_data(df dataframe.DataFrame) (dataframe.DataFrame, error) {
 	if df.Nrow() == 0 {
 		return df, fmt.Errorf("no valid records present in CSV to process further")
 	}
-
 	df = determine_k8s_object_type(df)
-
-	// filter out only valid workload type
-	df = filter_valid_k8s_object_types(df)
 
 	dfGroups := df.GroupBy(
 		"namespace",
@@ -85,8 +81,8 @@ func Aggregate_data(df dataframe.DataFrame) (dataframe.DataFrame, error) {
 	return df, nil
 }
 
-func filter_valid_csv_records(main_df dataframe.DataFrame) (dataframe.DataFrame, int) {
-	df := main_df.FilterAggregation(
+func filter_valid_csv_records(mainDf dataframe.DataFrame) (dataframe.DataFrame, int) {
+	df := mainDf.FilterAggregation(
 		dataframe.And,
 		dataframe.F{Colname: "memory_rss_usage_container_sum", Comparator: series.GreaterEq, Comparando: 0},
 		dataframe.F{Colname: "memory_rss_usage_container_max", Comparator: series.GreaterEq, Comparando: 0},
@@ -104,17 +100,34 @@ func filter_valid_csv_records(main_df dataframe.DataFrame) (dataframe.DataFrame,
 		dataframe.F{Colname: "owner_name", Comparator: series.Neq, Comparando: ""},
 		dataframe.F{Colname: "owner_kind", Comparator: series.Neq, Comparando: "<none>"},
 		dataframe.F{Colname: "owner_name", Comparator: series.Neq, Comparando: "<none>"},
+		dataframe.F{Colname: "workload_type", Comparator: series.Neq, Comparando: "<none>"},
+		dataframe.F{Colname: "workload_type", Comparator: series.Neq, Comparando: ""},
 	)
 
-	no_of_dropped_records := main_df.Nrow() - df.Nrow()
+	// The above filters can delete all the rows
+	if df.Nrow() == 0 {
+		return df, mainDf.Nrow()
+	}
 
-	return df, no_of_dropped_records
-}
+	// Change the case of all workload_type to lowercase
+	lcaseWorkloadTypes := df.Rapply(func(s series.Series) series.Series {
+		columns := df.Names()
+		indexOfWorkloadType := findInStringSlice("workload_type", columns)
+		workloadType := s.Elem(indexOfWorkloadType).String()
+		lcaseWorkloadType := strings.ToLower(workloadType)
+		return series.Strings([]string{lcaseWorkloadType})
+	})
 
-func filter_valid_k8s_object_types(df dataframe.DataFrame) dataframe.DataFrame {
-	return df.Filter(
+	// Delete existing workload_type column
+	df = df.Mutate(df.Col("workload_type")).Drop("workload_type")
+
+	// Rename lowercase converted column to workload_type
+	df = df.Mutate(lcaseWorkloadTypes.Col("X0")).Rename("workload_type", "X0")
+
+	df = df.FilterAggregation(
+		dataframe.And,
 		dataframe.F{
-			Colname:    "k8s_object_type",
+			Colname:    "workload_type",
 			Comparator: series.In,
 			Comparando: []string{
 				w.Daemonset.String(),
@@ -125,6 +138,10 @@ func filter_valid_k8s_object_types(df dataframe.DataFrame) dataframe.DataFrame {
 				w.Statefulset.String(),
 			}},
 	)
+
+	noOfDroppedRecords := mainDf.Nrow() - df.Nrow()
+
+	return df, noOfDroppedRecords
 }
 
 func determine_k8s_object_type(df dataframe.DataFrame) dataframe.DataFrame {
--- a/internal/utils/aggregator_test.go
+++ b/internal/utils/aggregator_test.go
@@ -1,7 +1,6 @@
 package utils
 
 import (
-	"fmt"
 	"testing"
 
 	"github.com/go-gota/gota/dataframe"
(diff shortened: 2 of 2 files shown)


upadhyeammit:
@saltgen as discussed I saw the potential to add this in `filter_valid_csv_records` however in this method we dont have the column `k8s_object_type` created and it happens in later in `determine_k8s_object_type`.

With the code change the processor does not panic and errors out,
~~~
ERRO[0005]/home/amit/work/insights/ros/ros-ocp-backend/internal/services/report_processor.go:80 github.com/redhatinsights/ros-ocp-backend/internal/services.ProcessReport() Error: no valid k8s_object_types present in CSV to process further  account= cluster_alias=name222 cluster_uuid=222 org_id=3340851 request_id=uuid1234 service=rosocp source_id=111
~~~

saltgen (changes requested):
@upadhyeammit As discussed, let's add the filter for workload type, on filter_csv_records.
This enum equivalent should help you, /internal/types/workload.

internal/utils/aggregator.go
```diff
@@ -86,6 +82,37 @@ func Aggregate_data(df dataframe.DataFrame) (dataframe.DataFrame, error) {
 }
 
 func filter_valid_csv_records(main_df dataframe.DataFrame) (dataframe.DataFrame, int) {
+
+	emptyWorkload := func() func(el series.Element) bool {
+		return func(el series.Element) bool {
+			if el.Type() == series.String {
+				if workload, ok := el.Val().(string); ok {
+					if workload == "<none>" || workload == "" {
```
upadhyeammit, line 90: The main branch code accepts `""` and `"<none>" workloads. I like to know should we filter it out then?

internal/utils/aggregator.go
```diff
@@ -104,17 +101,8 @@ func filter_valid_csv_records(main_df dataframe.DataFrame) (dataframe.DataFrame,
...
-	no_of_dropped_records := main_df.Nrow() - df.Nrow()
-
-	return df, no_of_dropped_records
-}
-
-func filter_valid_k8s_object_types(df dataframe.DataFrame) dataframe.DataFrame {
-	return df.Filter(
 		dataframe.F{
-			Colname:    "k8s_object_type",
+			Colname:    "workload_type",
```
patilsuraj767, line 105: Any specific reason we changed the colname from `k8s_object_type` to `workload_type`? 
As far as I remember `workload_type` does not tell the correct representation of workload. For example - even if the workload is of type replicaset, `workload_type` field is set to `deployment`.  Due to this behaviour we have a function `determine_k8s_object_type` which determines the correct type of workload and sets it in `k8s_object_type` column.

patilsuraj767 (dismissed):

internal/utils/aggregator.go
upadhyeammit: Actually let me explain more maybe,
The failure was happening because there were no rows which had valid `k8s_object_type` or I can say it was empty. However we did not had `df.Nrows==0` after filtering the valid `k8s_object_type` and so we ended up giving empty df to Groups further.
It looked we can just check `df.Nrows==0` after `filter_valid_k8s_object_types` and that should fix it, and it can. However then it also felt we can drop the empty and non valid rows in `filter_valid_csv_records` based on the workload_type, and then do the `determine_k8s_object_type` . So indeed we do not need to do `filter_valid_k8s_object_types` after `determine_k8s_object_type`?

patilsuraj767: Got your point. I agree we can apply filter before calling `determine_k8s_object_type`. But my concern is that you have removed the `filter_valid_k8s_object_types` function which actually filters workloads based on there workload_type. So in this case workload of type `job` will also get pass.

patilsuraj767: Instead of comparing actual workload type, can we simply add here `dataframe.F{Colname: "workload_type", Comparator: series.Neq, Comparando: ""},`

patilsuraj767: and keep `filter_valid_k8s_object_types` function as it is.

.~lock.output.csv#
saltgen, line 1: Let's remove this file

saltgen (changes requested):
Some changes were reviewed and suggestions provided over DM.

- var name of workloadType index
- single column for workloadType lowercase

internal/utils/aggregator_test.go
```diff
@@ -93,9 +120,8 @@ func Test_filter_valid_k8s_object_types(t *testing.T) {
 			"1", "1", "1", "1", "1", "1", "1", "1", "1", "1", "1", "1", "1", "1", "1", "1", "1", "1", "1", "1", "1", "1", "1",
 		},
 	}
-	df := dataframe.LoadStructs(usage_data)
-	df = determine_k8s_object_type(df)
-	result := filter_valid_k8s_object_types(df)
+	df = dataframe.LoadStructs(usage_data)
+	result, _ = filter_valid_csv_records(df)
 	fmt.Println(result.Nrow())
```
saltgen, line 125: Let's remove this statement

internal/utils/aggregator_test.go
saltgen, line 1: @upadhyeammit Since these changes cover the edge case being handled when workload and workload_name is empty, we should have a test case for the same, let's say after L#136?

saltgen (changes requested):

internal/utils/aggregator.go
```diff
@@ -104,27 +100,48 @@ func filter_valid_csv_records(main_df dataframe.DataFrame) (dataframe.DataFrame,
 		dataframe.F{Colname: "owner_name", Comparator: series.Neq, Comparando: ""},
 		dataframe.F{Colname: "owner_kind", Comparator: series.Neq, Comparando: "<none>"},
 		dataframe.F{Colname: "owner_name", Comparator: series.Neq, Comparando: "<none>"},
+		dataframe.F{Colname: "workload_type", Comparator: series.Neq, Comparando: "<none>"},
+		dataframe.F{Colname: "workload_type", Comparator: series.Neq, Comparando: ""},
 	)
 
-	no_of_dropped_records := main_df.Nrow() - df.Nrow()
+	// The above filters can delete all the rows
+	if df.Nrow() == 0 {
```
upadhyeammit, line 108: @saltgen writing more tests unearth this situation, added a code for same!

internal/utils/aggregator_test.go
```diff
@@ -178,3 +190,27 @@ func Test_check_if_all_required_columns_in_CSV(t *testing.T) {
 		t.Error("Expecting error to be returned as all required column not present")
 	}
 }
+
+func Test_Aggregate_data(t *testing.T) {
```
upadhyeammit, line 194: Added a test on Aggregate to attest condition which fixed this bug!

internal/utils/aggregator_test.go
```diff
@@ -112,43 +136,31 @@ func Test_filter_valid_k8s_object_types(t *testing.T) {
...
-func Test_filter_valid_csv_records(t *testing.T) {
-	usage_data := []UsageData{
-		// k8s object with missing data
-		{
-			"2023-02-01 00:00:00 +0000 UTC", "2023-03-01 00:00:00 +0000 UTC", "2023-06-02 00:00:01 +0000 UTC", "2023-06-02 00:15:00 +0000 UTC",
-			"Yuptoo-service", "Yuptoo-app-standalone-1", "Yuptoo-app", "ReplicaSet", "testdeployment", "deployment", "Yuptoo-prod",
-			"quay.io/cloudservices/yuptoo", "ip-10-0-176-227.us-east-2.compute.internal", "i-0dfbb3fa4d0e8fc94",
-			"1", "1", "1", "1", "1", "1", "1", "1", "1", "1", "1", "1", "1", "1", "1", "1", "1", "1", "1", "1", "1", "1", "",
-		},
+	// check if empty workload_type is dropped
```
saltgen, line 144: I don't think we'll need to check for the empty/`<none>` workloadType here, anymore.
This is because you have added the changes in `Test_Aggregate_data` that,

1. checks for empty, `<none>` workload_type
2. gracefully asserts that function does not panic

internal/utils/aggregator_test.go
saltgen:
```suggestion
func TestAggregateDataNoRecords(t *testing.T) {
```

saltgen (changes requested):

internal/utils/aggregator_test.go
saltgen:
```suggestion
func TestAggregateDataNoRecords(t *testing.T) {
```

internal/utils/aggregator.go
saltgen: Good Stuff! 🌟

internal/utils/aggregator_test.go
upadhyeammit: hmm, we have tests for most of the filters? That does mean in future if someone removes these filters for dropping empty `workload_type` then the test should fail?
The Test_aggregate_data is using same `usage_data` however intention is different, and its about `gracefully asserts that function does not panic`

saltgen (approved):
Thanks for the changes @upadhyeammit. Appreciate your involvement and approach! 🙌🏼

upadhyeammit:
> Thanks for the changes @upadhyeammit. Appreciate your involvement and approach! 🙌🏼

thanks for all the suggestions and help!

PR #324, opened 2025-03-18
RHINENG-16692 Omit pods_count field in recommendations

saltgen:

`pods_count` field/feature from Kruize is still at it's nascent stages

- [ ] Yes
- [x] No

## Security Checklist :lock:

- [ ] Bugfix
- [ ] New Feature
- [x] Refactor
- [ ] Unittests Added
- [ ] DRY code
- [ ] Dependency Added
- [ ] DB Migration Added

## Additional :mega:

`confidence_level` was already hidden from API responses due to the `omitempty` tag as Kruize has been responding with 0 for the field.

Removing the fields from respective struct is a standard way to ensure the `key;value` never gets captured and saved to database by ROS OCP.

### DIFF
--- a/internal/types/kruizePayload/common.go
+++ b/internal/types/kruizePayload/common.go
@@ -50,11 +50,14 @@ type Notification struct {
 }
 
 type RecommendationEngineObject struct {
-	PodsCount       int                     `json:"pods_count,omitempty"`
-	ConfidenceLevel float64                 `json:"confidence_level,omitempty"`
-	Config          ConfigObject            `json:"config,omitempty"`
-	Variation       ConfigObject            `json:"variation,omitempty"`
-	Notifications   map[string]Notification `json:"notifications"`
+	/*
+		Following fields/features from Kruize are at their nascent stages
+		PodsCount       int                     `json:"pods_count,omitempty"`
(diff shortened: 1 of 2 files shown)


internal/types/kruizePayload/common.go
```diff
@@ -50,11 +50,14 @@ type Notification struct {
...
-	PodsCount       int                     `json:"pods_count,omitempty"`
-	ConfidenceLevel float64                 `json:"confidence_level,omitempty"`
-	Config          ConfigObject            `json:"config,omitempty"`
-	Variation       ConfigObject            `json:"variation,omitempty"`
-	Notifications   map[string]Notification `json:"notifications"`
+	/*
+		Following fields/features from Kruize are at their nascent stages
+		PodsCount       int                     `json:"pods_count,omitempty"`
+		ConfidenceLevel float64                 `json:"confidence_level,omitempty"`
+	*/
```
upadhyeammit, line 57: Does this mean we are also not going to save these details in db?

saltgen: Yep

upadhyeammit (approved):
thanks for the change!

PR #412, opened 2025-07-01
RHINENG-18956 Support for csv format on API

saltgen:
### Support for csv format on API

- Added a query parameter `format`, options are json(default) and csv.
- A maximum of 1000 db records can be requested in csv format at a time. Each record can have three term recommendations, each of those term recommendations can have performance and cost recommendations, so a total of 6000 rows.
- The CSV response is supported only on the List endpoint.
- The CSV rows are streamed, once every CSV_STREAM_INTERVAL. This helps to avoid generating the complete csv in one go and in memory.
- Real-world units/format is available in the CSV.
- notification object rows are not present in the csv since the rows are already quite lengthy.

- [ ] Yes
- [x] No

## Security Checklist :lock:

- [ ] Bugfix
- [x] New Feature
- [ ] Refactor
- [ ] Unittests Added
- [ ] DRY code
- [ ] Dependency Added
- [ ] DB Migration Added

## Additional :mega:

Manual testing:
`curl -JO '/api/cost-management/v1/recommendations/openshift?start_date=1970-01-01&format=csv' -H "X-Rh-Identity: xxx"`

## Summary by Sourcery

Enable CSV export support on the recommendations list API by adding CSV generation and streaming with configurable db record limits and flush intervals.

New Features:
- Add CSV export support for the recommendations list endpoint via ?format=csv
- Stream CSV rows with periodic flushes to avoid high memory usage
- Enforce configurable db record limit (RecordLimitCSV) for CSV exports

Enhancements:
- Extract calculatePercentage into a standalone utility function
- Implement GenerateCSVRows and GenerateAndStreamCSV helpers for CSV row construction and streaming

Chores:
- Update GetRecommendationSets signature to accept format parameter and adjust limits for CSV
- Add FlattenedCSVHeader constant for CSV column names
- Introduce RecordLimitCSV and CSVStreamInterval configuration parameters with defaults


saltgen:
@sourcery-ai guide

saltgen:
@sourcery-ai review

Aurobinda55:
QE Updates : 
@saltgen, I went through the PR and checked the API, and here are my findings below. Also added the IQE test to cover the following things (see the PR https://gitlab.cee.redhat.com/insights-qe/iqe-ros-ocp-plugin/-/merge_requests/236 for reference )

- The recommendation API worked for both format=csv and json 
- Verified the content-type is text/csv and application/json for CSV and JSON format 
- Checked a few key fields and are expected in both types of responses

But I see some issues/ doubts  in the data that I got in the response  in both formats, 

- In the JSON response, the CPU limit format appears as an empty list, i.e., " ".  Should we consider adding "format": "cores" when the output format is JSON?

- There's an inconsistency between the memory limit representations in CSV and JSON formats:
CSV: 'current_memory_limit_amount': '3.221225472e+09'
JSON: 'memory': {'amount': 3221225472} .
Is this difference in representation expected?

- Some fields appear to be missing in the CSV-formatted data. For instance, the following entry is present in the JSON response but not in the CSV:


```
'memory': {'amount': 3605862601.8, 'format': 'bytes'},
'requests': {
    'cpu': {'amount': 2.24, 'format': ''},
    'memory': {'amount': 3605862601.8, 'format': 'bytes'}
}
```
I’ve attached a sample of both the CSV and JSON data for reference:
https://privatebin.corp.redhat.com/?0e4f9d63c2f36dba#Au9nnJsZFKX8T5GkwBuNJVa6TNuiuG6nSTok8qsJcRgc 

Also, I noticed the CSV filename is included in the response headers, but the file doesn’t seem to be saved or accessible at the expected path. Is there a way to get the file?

For example, in the following output:


```
In [13]: response_csv.headers 
Out[13]: HTTPHeaderDict({
    'Content-Disposition': 'attachment; filename=recommendations-20250701.csv',
    'Content-Type': 'text/csv',
    'Vary': 'Origin',
    'Date': 'Tue, 01 Jul 2025 15:23:47 GMT',
    'Transfer-Encoding': 'chunked'
})
```
I'm unable to locate the file in the expected download path.

saltgen:
> QE Updates : @saltgen, I went through the PR and checked the API, and here are my findings below. Also added the IQE test to cover the following things (see the PR https://gitlab.cee.redhat.com/insights-qe/iqe-ros-ocp-plugin/-/merge_requests/236 for reference )
> 
> * The recommendation API worked for both format=csv and json
> * Verified the content-type is text/csv and application/json for CSV and JSON format
> * Checked a few key fields and are expected in both types of responses
> 
> But I see some issues/ doubts in the data that I got in the response in both formats,
> 
> * In the JSON response, the CPU limit format appears as an empty list, i.e., " ".  Should we consider adding "format": "cores" when the output format is JSON?
> * There's an inconsistency between the memory limit representations in CSV and JSON formats:
>   CSV: 'current_memory_limit_amount': '3.221225472e+09'
>   JSON: 'memory': {'amount': 3221225472} .
>   Is this difference in representation expected?
> * Some fields appear to be missing in the CSV-formatted data. For instance, the following entry is present in the JSON response but not in the CSV:
> 
> ```
> 'memory': {'amount': 3605862601.8, 'format': 'bytes'},
> 'requests': {
>     'cpu': {'amount': 2.24, 'format': ''},
>     'memory': {'amount': 3605862601.8, 'format': 'bytes'}
> }
> ```
> 
> I’ve attached a sample of both the CSV and JSON data for reference: https://privatebin.corp.redhat.com/?0e4f9d63c2f36dba#Au9nnJsZFKX8T5GkwBuNJVa6TNuiuG6nSTok8qsJcRgc
> 
> Also, I noticed the CSV filename is included in the response headers, but the file doesn’t seem to be saved or accessible at the expected path. Is there a way to get the file?
> 
> For example, in the following output:
> 
> ```
> In [13]: response_csv.headers 
> Out[13]: HTTPHeaderDict({
>     'Content-Disposition': 'attachment; filename=recommendations-20250701.csv',
>     'Content-Type': 'text/csv',
>     'Vary': 'Origin',
>     'Date': 'Tue, 01 Jul 2025 15:23:47 GMT',
>     'Transfer-Encoding': 'chunked'
> })
> ```
> 
> I'm unable to locate the file in the expected download path.

1. "" is kubernetes representation of `cores`, similarly `MiB` is represented as "Mi" in k8s, for the csv real-world units make sense, on the JSON however these configs can be copy pasted by users hence the k8s representation adherence.

2. The exponential notation I believe is being done by the app, Google sheets in this case. I think this should be fine.
image

3. Each term recommendation will have 2 rows, cost and performance. From the snippet, the CSV row seems to be of cost type, the value in Line 24 is from the performance row, could you check if the performance row is present for the same recommendation? Additionally I have skipped csv rows for notification objects, since the rows are already quite lengthy.

4. You'll need to pass the `-JO` flag with curl such that the file gets saved with the name present in the header, please check the Additional section of this PR.

P.S. 1 and 3 should have been updated in the PR description, I will add them now.

saltgen:
@r14chandra Added a test

r14chandra (changes requested):
Added a few comments.

internal/api/utils.go
```diff
@@ -610,3 +615,136 @@ func UpdateRecommendationJSON(handlerName string, recommendationID string, clust
...
+		if generateRowErr != nil {
+			return fmt.Errorf("unable to generate rows: %w", generateRowErr)
+		}
+		for _, row := range CSVRows {
+			if err := writer.Write(row); err != nil {
+				return fmt.Errorf("unable to write row: %w", err)
+			}
+		}
+
+		if (i+1)%config.GetConfig().CSVStreamInterval == 0 { // flush every CSVStreamInterval rows
```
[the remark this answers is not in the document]
saltgen: The original approach was to flush based on a preset number of db records.
For instance, we have 100 as CSV_STREAM_INTERVAL, so the flush will happen for every 100*3(term)*2(engine) i.e. 600 rows.
The API service might send rows beyond 600 records because not every db record will have short, medium and long term records.
That said each term > engine recomm. row will be present per record because we are flushing by db records.
Let me know if you see any benefits we are missing out on by taking this approach.

saltgen:
@r14chandra All suggested changes pushed

Aurobinda55:
@saltgen The changes look good to me from the QE side of verifications. The IQE pR is updated with the new verification checks that we discussed over the call. Please take a look at the PR and let me know if anything you want to discuss. Thanks!

saltgen:
> @saltgen The changes look good to me from the QE side of verifications. The IQE pR is updated with the new verification checks that we discussed over the call. Please take a look at the PR and let me know if anything you want to discuss. Thanks!

@Aurobinda55 IQE tests look good, PR acked

internal/api/api_test.go
```diff
@@ -74,3 +75,50 @@ func TestMapQueryParameters(t *testing.T) {
 		})
 	}
 }
+
+var FlattenedCSVHeaderFixture = []string{
```
kgaikwad, line 79: @saltgen - I noticed that the same constant variable is declared separately in both the App level and the tests. While this works, it can lead to duplication and maintenance issues if the constant ever needs to change. 
Do you think that it might be cleaner and safer to define the constant once in a shared package and reuse it in both the app and tests? (example - we already have api package > common file OR one can declare specific to CSV) This keeps things DRY and ensures consistency across the codebase.

r14chandra (approved):
The PR looks good to me! 🚀

openapi.json
```diff
@@ -23,6 +23,17 @@
         },
         "operationId": "getRecommendationList",
         "parameters": [
+          {
```
kgaikwad, line 26: I am not sure I understand the need for introducing an additional request parameter to control the response format. Could we instead rely on the standard Accept header to determine whether the client expects JSON or CSV? This would follow HTTP best practices (https://developer.mozilla.org/en-US/docs/Web/HTTP/Reference/Headers/Accept) and simplify the API surface for users.

The handler can inspect the Accept header and respond accordingly by setting the appropriate `Content-Type` (which you already set). Is there a specific reason we need the custom parameter?

openapi.json
```diff
@@ -23,6 +23,17 @@
...
+            "name": "format",
+            "in": "query",
+            "description": "Maximum number of records is 1000 i.e. 6000 rows for CSV downloads. The 'offset' parameter can be used for pagination with both formats.",
```
kgaikwad, line 29: To describe, request or response you can make use of Header Parameters for openapi file. 
Please refer below links:
Header parameter - https://swagger.io/docs/specification/v3_0/describing-parameters/#header-parameters
Response Media Types - https://swagger.io/docs/specification/v3_0/describing-responses/#response-media-types

kgaikwad (changes requested):

internal/api/api_test.go
saltgen: @kgaikwad The test and app should ideally have separate objects, the one in the test is suffixed with `Fixture`.
Tests should have separate context altogether.

openapi.json
saltgen: It is a common practice to use query parameters for such use cases, because it provides a direct link for the user `/data?format=csv`.
Additionally, the Accept header is useful for direct API users, whereas for browsers and simple clients the query parameters are a bit more straight-forward.
I think we should add the Accept header as an enhancement, could you create a card for the same and assign it to me?

PR #419, opened 2025-07-10
RHINENG-19228 `Accept` header with query param `format`

saltgen:

Accept header takes precendence over format query param
both `accept` and `Accept` can be interpreted by the service

- [ ] Yes
- [ ] No

## Security Checklist :lock:

- [ ] Bugfix
- [ ] New Feature
- [ ] Refactor
- [ ] Unittests Added
- [ ] DRY code
- [ ] Dependency Added
- [ ] DB Migration Added

## Additional :mega:

### DIFF
--- a/internal/api/handlers.go
+++ b/internal/api/handlers.go
@@ -111,19 +111,31 @@ func GetRecommendationSetList(c echo.Context) error {
 		}
 	}
 
-	format := c.QueryParam("format")
+	acceptHeaderValue := c.Request().Header.Get("Accept")
+	format := "json" // default value
 
-	if format == "" {
-		format = "json"
-	} else {
-		formatLower := strings.ToLower(format)
-		switch formatLower {
-		case "csv":
-			format = "csv"
-		case "json":
-			format = "json"
-		default:
-			return c.JSON(http.StatusBadRequest, echo.Map{"status": "error", "message": "invalid value for format"})
+	switch acceptHeaderValue {
+	case "text/csv":
+		format = "csv"
+	case "application/json":
+		break
+	default:
+		formatParam := c.QueryParam("format")
+		if formatParam == "" {
+			break
+		} else {
+			formatLower := strings.ToLower(formatParam)
+			switch formatLower {
+			case "csv":
+				format = "csv"
+			case "json":
+				break
+			default:
+				format = formatLower
+				return c.JSON(http.StatusBadRequest,
+					echo.Map{"status": "error", "message": fmt.Sprintf("invalid value for format; %s", format)},
+				)
+			}
 		}
 	}
 


saltgen:
@sourcery-ai review

internal/api/utils.go
```diff
@@ -617,8 +617,13 @@ func UpdateRecommendationJSON(handlerName string, recommendationID string, clust
...
 func formatPrecisionValuesToStr(val float64) string {
-	s := fmt.Sprintf("%.3f", val)
-	s = strings.TrimRight(s, "0") // removes trailing zeros
+	// avoid un-necessary rounding by sprintf 104.939886 -> -104.940
+	multiplier := 1000.0
+	truncatedVal := math.Trunc(val*multiplier) / multiplier
+
+	s := fmt.Sprintf("%.3f", truncatedVal)
+	s = strings.TrimRight(s, "0")  // removes trailing zeros
+	s = strings.TrimSuffix(s, ".") // removes trailing "." for instance 10.0
```
saltgen, line 626: @kgaikwad Since this is a minor but indirectly related change I have pushed the code in this PR.
Let me know if you think we should handle this separately, thanks.

Aurobinda55:
> @Aurobinda55 I see an AssertionError: assert 'application/json' == 'text/csv' for all the failed iqe tests, by any chance does `get_response_and_data("csv")` set `Accept: application/json`? If that is the case the `format` query parameter will be overridden by accept header value.

@saltgen  Yes, I see the request contains `Accept: application/json` by default in every request call. But the tests are not always failing; sometimes they fail with the assertion error.  I am looking into it.

internal/api/utils.go
```diff
@@ -752,3 +757,32 @@ func GenerateAndStreamCSV(w io.Writer, recommendationSets []model.Recommendation
 	}
 	return nil
 }
+
+func resolveResponseFormat(acceptHeaderVal string, formatQueryParamVal string) (string, error) {
```
kgaikwad, line 761: From this method, it looks like, you are only consider `format` query param value when `Accept` header is not set or missing or not supported. correct?
That means  `format` query param is being used as fallback and not to override the value whereas in few of the APIs implementation side, query parameters are allowed to override headers explicitly. So, in my opinion, you should clearly document this here - https://github.com/RedHatInsights/ros-ocp-backend/blob/main/openapi.json#L26 to clarify the rationale for including the format query parameter.

internal/api/utils.go
```diff
@@ -752,3 +757,32 @@ func GenerateAndStreamCSV(w io.Writer, recommendationSets []model.Recommendation
...
+	switch acceptHeaderVal {
+	case "text/csv":
+		responseFormat = "csv"
+	case "application/json":
+		responseFormat = "json"
+	}
+
+	if responseFormat != "" {
+		return responseFormat, nil // preferring header value
+	} else {
```
kgaikwad, line 776: What is the expected behavior here when someone passes accept header's value other than `text/csv` or `application/json` and `format` as `json` or `csv` ?  

@Aurobinda55  do you think this scenario should also be captured in a test?

kgaikwad (changes requested):

internal/api/utils.go
saltgen: The `accept/Accept` header value will checked first if value is invalid/empty, the format param value will be evaluated.
Let's say request has `"Accept": "application/json"` + format=csv, response will be provided in JSON.
Documented, [here](https://github.com/RedHatInsights/ros-ocp-backend/pull/419/files#diff-a9865368a7fc7fa33065e35b2343f10d08fb79d65205435403d0a163a3044713R29).

openapi.json
```diff
@@ -26,7 +26,7 @@
           {
             "name": "format",
             "in": "query",
-            "description": "Maximum number of records is 1000 i.e. 6000 rows for CSV downloads. The 'offset' parameter can be used for pagination with both formats.",
+            "description": "Ignored when 'Accept' header is present and valid.  \nMaximum number of records is 1000 i.e. 6000 rows for CSV downloads.  \nThe 'offset' parameter can be used for pagination with both formats.",
```
kgaikwad, line 29:
```suggestion
            "description": "Used as a fallback when the 'Accept' header is missing or specifies an unsupported media type.  \nMaximum number of records is 1000 i.e. 6000 rows for CSV downloads.  \nThe 'offset' parameter can be used for pagination with both formats.",
```

In my opinion, ROS OCP supports both JSON and CSV formats, but that doesn't mean other media types are invalid. As per [MDN's list of common MIME types](https://developer.mozilla.org/en-US/docs/Web/HTTP/Guides/MIME_types/Common_types), formats like text/css are valid but it is not applicable to endpoint of  ROS OCP service. It would be more accurate to refer to them as "unsupported" rather than "invalid."

Thoughts?

kgaikwad (approved):
LGTM :+1:

PR #618, opened 2026-04-08
Fixes #RHINENG-25638 - Store and use percentage variations for sorting

upadhyeammit:
## PR Title :boom:

Suggested formats: 

2. RHIROS-XXX Title
3. RHINENG-XXX Title 

With this we store the actual percentage variation on db table for sorting so final sorted results on API are sorted correctly.

- [ ] Yes
- [ ] No

## Security Checklist :lock:

- [ ] Bugfix
- [ ] New Feature
- [ ] Refactor
- [ ] Unittests Added
- [ ] DRY code
- [ ] Dependency Added
- [ ] DB Migration Added

## Additional :mega:

## Summary by Sourcery

Store namespace recommendation variation values as percentage-of-request and align sorting with API semantics while tightening ORDER BY validation for namespace lists.

New Features:
- Persist per-term CPU and memory variation percentages for namespace recommendations to support consistent sorting with API responses.

Bug Fixes:
- Fix incorrect namespace recommendation sorting by using percentage-based variation values instead of raw amounts.
- Prevent invalid ORDER BY and ORDER HOW values for namespace recommendation list queries to avoid unsafe or inconsistent sorting.

Enhancements:
- Refactor percentage calculation into a shared utility and reuse it across recommendation extraction, API JSON transformation, and CSV generation.
- Add deterministic secondary sorting by namespace recommendation set ID for stable ordered results.

Chores:
- Introduce a DB migration replacing legacy raw variation columns with explicit *_pct percentage columns for namespace recommendation sets.


migrations/000022_namespace_variation_columns_percent_replace.up.sql
r14chandra, line 1: As Sourcery highlighted about data loss when current columns are dropped. How would it be if we create the new column first, add existing values after conversion, then drop the old columns? Same for down migration.

r14chandra (changes requested):

r14chandra (commented):
There is a linting error

r14chandra:
@sourcery-ai review

migrations/000022_namespace_variation_columns_percent_replace.up.sql
upadhyeammit: These new columns are not in production yet so ideally there is nothing as data loss. At the same time, data loss would be meaningful if we dont have that data at all, however these columns are derived from recommendation json and this data is already available.

internal/utils/conversion.go
saltgen, line 1: Could you share the context behind relocating the percentage util func?

internal/api/listoptions/list_options.go
```diff
@@ -50,19 +50,42 @@ var NsAllowedOrderBy = OrderByMap{
...
+// ValidNamespaceListOrderColumn reports whether columnSQL is an allowed ORDER BY expression for namespace lists.
+// Use before building GORM Order() for namespace recommendations (defense in depth vs untrusted ListOptions).
+func ValidNamespaceListOrderColumn(columnSQL string) bool {
+	_, ok := allowedNamespaceOrderBySQL[columnSQL]
+	return ok
+}
+
+// ValidListOrderHow reports whether how is asc or desc only.
+func ValidListOrderHow(how string) bool {
+	return how == OrderAsc || how == OrderDesc
```
saltgen: I find these validators to be redundant, the explicit map allowed options = db.column already provides validation and guards against SQL injection.

P.S. This change does not seem to get rid of the Code QL warning as well. CodeQL is unable to model the relationship between ValidNamespaceListOrderColumn returning true and opts.OrderBy being safe, this is a static linter limitation and can be ignored.

internal/api/listoptions/list_options.go
```diff
@@ -50,19 +50,42 @@ var NsAllowedOrderBy = OrderByMap{
...
+	"cpu_variation_medium_cost_pct":           "namespace_recommendation_sets.cpu_variation_medium_cost_pct",
+	"cpu_variation_medium_performance_pct":    "namespace_recommendation_sets.cpu_variation_medium_performance_pct",
+	"cpu_variation_long_cost_pct":             "namespace_recommendation_sets.cpu_variation_long_cost_pct",
+	"cpu_variation_long_performance_pct":      "namespace_recommendation_sets.cpu_variation_long_performance_pct",
+	"memory_variation_short_cost_pct":         "namespace_recommendation_sets.memory_variation_short_cost_pct",
+	"memory_variation_short_performance_pct":  "namespace_recommendation_sets.memory_variation_short_performance_pct",
+	"memory_variation_medium_cost_pct":        "namespace_recommendation_sets.memory_variation_medium_cost_pct",
+	"memory_variation_medium_performance_pct": "namespace_recommendation_sets.memory_variation_medium_performance_pct",
+	"memory_variation_long_cost_pct":          "namespace_recommendation_sets.memory_variation_long_cost_pct",
+	"memory_variation_long_performance_pct":   "namespace_recommendation_sets.memory_variation_long_performance_pct",
```
saltgen: I find the `_pct` in the key to be a bit verbose when thinking from the user/client side.
We can have something like so, and in the openapi file we can mention the description as percentage and type as float.

```
"memory_variation_long_performance":   "namespace_recommendation_sets.memory_variation_long_performance_pct"
```

migrations/000022_namespace_variation_columns_percent_replace.up.sql
```diff
@@ -0,0 +1,40 @@
...
+    ADD COLUMN cpu_variation_short_performance_pct FLOAT,
+    ADD COLUMN cpu_variation_medium_cost_pct FLOAT,
+    ADD COLUMN cpu_variation_medium_performance_pct FLOAT,
+    ADD COLUMN cpu_variation_long_cost_pct FLOAT,
+    ADD COLUMN cpu_variation_long_performance_pct FLOAT,
+    ADD COLUMN memory_variation_short_cost_pct FLOAT,
+    ADD COLUMN memory_variation_short_performance_pct FLOAT,
+    ADD COLUMN memory_variation_medium_cost_pct FLOAT,
+    ADD COLUMN memory_variation_medium_performance_pct FLOAT,
+    ADD COLUMN memory_variation_long_cost_pct FLOAT,
```
saltgen: The migration currently uses FLOAT (which is double precision in Postgres -- 15+ significant digits, no constraint). If an application bug skips truncation and writes something like 33.33333333333333, the DB would simply store it.

Using NUMERIC(10,4) instead would:

Enforce 3 decimal places at the DB level
Postgres rounds on insert, so a buggy write of 33.33339 becomes 33.333
Makes the schema self-documenting about expected precision.
Setting the bound as 4 would also help with order items precisely.

internal/model/common.go
```diff
@@ -48,23 +48,26 @@ func ExtractRecommendationColumnValues(data kruizePayload.RecommendationData) Re
...
 	}
 	if e := terms.Long_term.RecommendationEngines; e != nil {
-		recommVals.CPUVariationLongCost = e.Cost.Variation.Requests.Cpu.Amount
-		recommVals.MemoryVariationLongCost = e.Cost.Variation.Requests.Memory.Amount
-		recommVals.CPUVariationLongPerformance = e.Performance.Variation.Requests.Cpu.Amount
-		recommVals.MemoryVariationLongPerformance = e.Performance.Variation.Requests.Memory.Amount
+		recommVals.CPUVariationLongCostPct = utils.CalculatePercentage(e.Cost.Variation.Requests.Cpu.Amount, cpuReq)
+		recommVals.MemoryVariationLongCostPct = utils.CalculatePercentage(e.Cost.Variation.Requests.Memory.Amount, memReq)
+		recommVals.CPUVariationLongPerformancePct = utils.CalculatePercentage(e.Performance.Variation.Requests.Cpu.Amount, cpuReq)
+		recommVals.MemoryVariationLongPerformancePct = utils.CalculatePercentage(e.Performance.Variation.Requests.Memory.Amount, memReq)
```
saltgen: We are truncating cpu to 3 places and memory to 2 places before calculating percentages.

We will need to replicate the same behaviour here as well, else this would lead to such issues.

Memory example (2-decimal MiB truncation):

Row A -- variation: 1,152,385 bytes, current: 3,146,804 bytes

Raw: 1.099 MiB / 3.001 MiB
After truncation: 1.09 MiB / 3.00 MiB
DB percentage: 36.62%
API percentage: 36.33%
Row B -- variation: 1,143,997 bytes, current: 3,144,707 bytes

Raw: 1.091 MiB / 2.999 MiB
After truncation: 1.09 MiB / 2.99 MiB
DB percentage: 36.38%
API percentage: 36.45%
Result:

DB sorts A above B (36.62 > 36.38)
API shows B above A (36.45 > 36.33)
Sort order is flipped. User sees B has a higher percentage but it appears below A.

r14chandra:
@sourcery-ai review

saltgen (changes requested):

migrations/000022_namespace_variation_columns_percent_replace.up.sql
```diff
@@ -0,0 +1,40 @@
...
+ALTER TABLE namespace_recommendation_sets
+    DROP COLUMN IF EXISTS cpu_variation_short_cost,
+    DROP COLUMN IF EXISTS cpu_variation_short_performance,
+    DROP COLUMN IF EXISTS cpu_variation_medium_cost,
+    DROP COLUMN IF EXISTS cpu_variation_medium_performance,
+    DROP COLUMN IF EXISTS cpu_variation_long_cost,
+    DROP COLUMN IF EXISTS cpu_variation_long_performance,
+    DROP COLUMN IF EXISTS memory_variation_short_cost,
+    DROP COLUMN IF EXISTS memory_variation_short_performance,
+    DROP COLUMN IF EXISTS memory_variation_medium_cost,
```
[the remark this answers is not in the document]
r14chandra: Not relevant as per comment - [https://github.com/RedHatInsights/ros-ocp-backend/pull/618#discussion_r3051851266](https://github.com/RedHatInsights/ros-ocp-backend/pull/618#discussion_r3051851266)

migrations/000022_namespace_variation_columns_percent_replace.up.sql
upadhyeammit: Good thought and some of the questions,

Range: NUMERIC(10,4) allows up to 6 digits before the decimal (e.g. 999999.9999). Confirm that’s enough for your worst-case percent (e.g. large multipliers or sentinel values).

We store memory in bytes, its possible that we get the difference which is bigger than what is allowed here?

App stack: Go code uses float64 on the model today. NUMERIC still often maps to float64 via database/sql, but some setups prefer decimal types for money/percent to avoid any driver rounding at the edges. Worth a quick check with your actual driver/GORM usage for read/write of these columns.

I think unless we exactly know the bounds we should not change this?

saltgen: Yes, values in bytes can be super long and un-predictable however since we are going for % values I am thinking up to 10 digits with 4 places decimal should be enough.
We can higher too but I think it would be wise to set an upper bound.

internal/utils/conversion.go
upadhyeammit: thought to have separate dedicated util as I am touching the code instead of putting everything in common utils

Aurobinda55:
QE updates - 

- Checked the new variation fields - api now able to sort based on the percentage value and working as expected, also checked it uses the variation string without  "_pct" in the api
- Checked regression - All existing namespacerelated tests working fine.
- Added automation for the same (in review)

Overall looks good to me (checked on https://github.com/RedHatInsights/ros-ocp-backend/pull/618/commits/04fbbbaca7a0d8f00de889ea49b8c0638f6cf60b )

internal/api/listoptions/list_options.go
```diff
@@ -50,19 +50,32 @@ var NsAllowedOrderBy = OrderByMap{
...
```
saltgen: We don't require these, since these changes aren't on Prod.

migrations/000022_namespace_variation_columns_percent_replace.up.sql
```diff
@@ -0,0 +1,40 @@
...
+    DROP COLUMN IF EXISTS cpu_variation_medium_cost_pct,
+    DROP COLUMN IF EXISTS cpu_variation_medium_performance_pct,
+    DROP COLUMN IF EXISTS cpu_variation_long_cost_pct,
+    DROP COLUMN IF EXISTS cpu_variation_long_performance_pct,
+    DROP COLUMN IF EXISTS memory_variation_short_cost_pct,
+    DROP COLUMN IF EXISTS memory_variation_short_performance_pct,
+    DROP COLUMN IF EXISTS memory_variation_medium_cost_pct,
+    DROP COLUMN IF EXISTS memory_variation_medium_performance_pct,
+    DROP COLUMN IF EXISTS memory_variation_long_cost_pct,
+    DROP COLUMN IF EXISTS memory_variation_long_performance_pct;
```
saltgen: These *_pct columns are not present on `main` and can be safely removed.

internal/api/utils.go
```diff
@@ -892,11 +864,11 @@ func convertVariationToPercentage(recommendationJSON map[string]interface{}) map
 								if memoryValue, ok := memoryObject["amount"].(float64); ok {
 									switch section {
 									case "limits":
-										percentageMemoryValue := calculatePercentage(memoryValue, currentMemoryLimits)
-										memoryObject["amount"] = truncateToThreeDecimalPlaces(percentageMemoryValue)
+										percentageMemoryValue := utils.CalculatePercentage(memoryValue, currentMemoryLimits)
+										memoryObject["amount"] = utils.TruncateToThreeDecimalPlaces(percentageMemoryValue)
 									case "requests":
-										percentageMemoryValue := calculatePercentage(memoryValue, currentMemoryRequests)
-										memoryObject["amount"] = truncateToThreeDecimalPlaces(percentageMemoryValue)
```
saltgen, line 899: Since the variation values are being calculated and stored in db columns, I think it would be best to refer those from the same, instead of recalculating the same while preparing the api response.

Update: That said let's handle this while adding these changes for the container flow.

saltgen (changes requested):

internal/api/utils.go
upadhyeammit: Yeah, that was intention to handle that in next pr so we will have less code moving in and out.

saltgen (approved):
Thanks @upadhyeammit for the changes and revisions!

internal/utils/conversion.go
r14chandra, line 1: Should we have other naming for this file, as it also contains some truncation helpers? Maybe `conversion.go` or anything else

r14chandra (approved):
Looks good. Just added one comment

PR #626, opened 2026-04-17
Fixes #RHINENG-20861 - variation API changes for container and namespace

upadhyeammit:
## PR Title :boom:

Suggested formats: 

2. RHIROS-XXX Title
3. RHINENG-XXX Title 

Please include the __context of this change__ here. If the change depends on Kruize add details about that as well!

- [ ] Yes
- [ ] No

## Security Checklist :lock:

- [ ] Bugfix
- [ ] New Feature
- [ ] Refactor
- [ ] Unittests Added
- [ ] DRY code
- [ ] Dependency Added
- [ ] DB Migration Added

## Additional :mega:

## Summary by Sourcery

Store pre-computed variation percentages for container and namespace recommendations and use them to drive API responses and sorting.

New Features:
- Expose new ordering options on container recommendations for current CPU/memory requests and per-term CPU/memory variation percentages.

Enhancements:
- Embed stored variation percentage fields in recommendation result models and wire them into list and detail handlers for API response shaping.
- Extend recommendation poller to persist current request values and per-term CPU/memory variation percentages into recommendation_sets.
- Adjust variation-to-percentage transformation to optionally reuse stored request variation percentages rather than recomputing from JSON.
- Update SQL list options and tests so container recommendation queries sort on the new *_pct and current request columns.

Tests:
- Expand list options tests to cover ordering by container current requests and variation percentage columns.

Chores:
- Add a database migration to introduce current CPU/memory request columns on recommendation_sets with rollback support.

### DIFF
--- a/internal/api/handlers.go
+++ b/internal/api/handlers.go
@@ -53,6 +53,7 @@ func GetRecommendationSetList(c echo.Context) error {
 			unitChoices,
 			setk8sUnits,
 			recommendationSets[i].Recommendations,
+			&recommendationSets[i].StoredVariationPcts,
 		)
 	}
 
@@ -122,7 +123,9 @@ func GetRecommendationSet(c echo.Context) error {
 			recommendationSet.ClusterUUID,
 			unitChoices,
 			setk8sUnits,
-			recommendationSet.Recommendations)
+			recommendationSet.Recommendations,
+			nil, // single-item fetch: recompute variation percentages from JSON
+		)
 		return c.JSON(http.StatusOK, recommendationSet)
 	} else {
 		return c.JSON(http.StatusNotFound, echo.Map{"status": "not_found", "message": "recommendation not found"})
@@ -175,6 +178,7 @@ func GetNamespaceRecommendationSetList(c echo.Context) error {
 			unitChoices,
 			setk8sUnits,
 			namespaceRecommendationSets[i].Recommendations,
+			&namespaceRecommendationSets[i].StoredVariationPcts,
 		)
 	}
 
@@ -232,7 +236,9 @@ func GetNamespaceRecommendationSet(c echo.Context) error {
 			nsRecommendationSet.ClusterUUID,
 			unitChoices,
 			setk8sUnits,
-			nsRecommendationSet.Recommendations)
+			nsRecommendationSet.Recommendations,
+			nil, // single-item fetch: recompute variation percentages from JSON
+		)
 		return c.JSON(http.StatusOK, nsRecommendationSet)
 	} else {
 		return c.JSON(http.StatusNotFound, echo.Map{"status": "not_found", "message": "project recommendation not found"})
--- a/internal/api/handlers.go
+++ b/internal/api/handlers.go
@@ -57,6 +57,7 @@ func GetRecommendationSetList(c echo.Context) error {
 			unitChoices,
 			setk8sUnits,
 			recommendationSets[i].Recommendations,
+			&recommendationSets[i].StoredVariationPcts,
 		)
 	}
 
@@ -126,7 +127,9 @@ func GetRecommendationSet(c echo.Context) error {
 			recommendationSet.ClusterUUID,
 			unitChoices,
 			setk8sUnits,
-			recommendationSet.Recommendations)
+			recommendationSet.Recommendations,
+			nil, // single-item fetch: recompute variation percentages from JSON
+		)
 		return c.JSON(http.StatusOK, recommendationSet)
 	} else {
 		return c.JSON(http.StatusNotFound, echo.Map{"status": "not_found", "message": "recommendation not found"})
@@ -183,6 +186,7 @@ func GetNamespaceRecommendationSetList(c echo.Context) error {
 			unitChoices,
 			setk8sUnits,
 			namespaceRecommendationSets[i].Recommendations,
+			&namespaceRecommendationSets[i].StoredVariationPcts,
 		)
 	}
 
@@ -240,7 +244,9 @@ func GetNamespaceRecommendationSet(c echo.Context) error {
 			nsRecommendationSet.ClusterUUID,
 			unitChoices,
 			setk8sUnits,
-			nsRecommendationSet.Recommendations)
+			nsRecommendationSet.Recommendations,
+			nil, // single-item fetch: recompute variation percentages from JSON
+		)
 		return c.JSON(http.StatusOK, nsRecommendationSet)
 	} else {
 		return c.JSON(http.StatusNotFound, echo.Map{"status": "not_found", "message": "project recommendation not found"})
--- a/internal/api/handlers.go
+++ b/internal/api/handlers.go
@@ -128,7 +128,7 @@ func GetRecommendationSet(c echo.Context) error {
 			unitChoices,
 			setk8sUnits,
 			recommendationSet.Recommendations,
-			nil, // single-item fetch: recompute variation percentages from JSON
+			&recommendationSet.StoredVariationPcts,
 		)
 		return c.JSON(http.StatusOK, recommendationSet)
 	} else {
@@ -245,7 +245,7 @@ func GetNamespaceRecommendationSet(c echo.Context) error {
 			unitChoices,
 			setk8sUnits,
 			nsRecommendationSet.Recommendations,
-			nil, // single-item fetch: recompute variation percentages from JSON
+			&nsRecommendationSet.StoredVariationPcts,
 		)
 		return c.JSON(http.StatusOK, nsRecommendationSet)
 	} else {
--- a/internal/api/utils.go
+++ b/internal/api/utils.go
@@ -747,9 +747,12 @@ func dropBoxPlotsObject(recommendationJSON map[string]interface{}) map[string]in
 	return recommendationJSON
 }
 
-func convertVariationToPercentage(recommendationJSON map[string]interface{}) map[string]interface{} {
+// convertVariationToPercentage replaces variation amounts in the recommendation JSON with
+// percentages relative to the corresponding current amounts. When skipRequests is true the
+// "requests" section is left untouched (used when stored *_pct values have already been
+// injected via injectStoredRequestVariationPct).
+func convertVariationToPercentage(recommendationJSON map[string]interface{}, skipRequests bool) map[string]interface{} {
 	var currentCpuLimits, currentMemoryLimits, currentCpuRequests, currentMemoryRequests float64
-	// Current section of recommendation
 
 	current_config, ok := recommendationJSON["current"].(map[string]interface{})
 	if !ok {
@@ -810,7 +813,12 @@ func convertVariationToPercentage(recommendationJSON map[string]interface{}) map
 						continue
 					}
 
-					for _, section := range []string{"limits", "requests"} {
+					sections := []string{"limits", "requests"}
+					if skipRequests {
+						sections = []string{"limits"}
+					}
+
+					for _, section := range sections {
 						sectionObject, ok := recommendationSection[section].(map[string]interface{})
 						if ok {
 							memoryObject, ok := sectionObject["memory"].(map[string]interface{})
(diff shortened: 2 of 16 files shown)


internal/api/utils.go
```diff
@@ -750,9 +750,12 @@ func dropBoxPlotsObject(recommendationJSON map[string]interface{}) map[string]in
 	return recommendationJSON
 }
 
-func convertVariationToPercentage(recommendationJSON map[string]interface{}) map[string]interface{} {
+// convertVariationToPercentage replaces variation amounts in the recommendation JSON with
+// percentages relative to the corresponding current amounts. When skipRequests is true the
+// "requests" section is left untouched (used when stored *_pct values have already been
+// injected via injectStoredRequestVariationPct).
+func convertVariationToPercentage(recommendationJSON map[string]interface{}, skipRequests bool) map[string]interface{} {
```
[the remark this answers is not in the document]
upadhyeammit: This is good suggestion however we want to leverage using computed percentage values first instead of recalculating it. Later on the logic to calculate the percentages for requests can be removed once we have all the percentage values in database.

internal/model/common.go
```diff
@@ -15,10 +15,64 @@ const (
...
+func (s *StoredVariationPcts) Lookup(term, engine string) (cpu, mem *float64) {
+	switch {
+	case term == "short_term" && engine == "cost":
+		return s.CPUVariationShortCostPct, s.MemoryVariationShortCostPct
+	case term == "short_term" && engine == "performance":
+		return s.CPUVariationShortPerformancePct, s.MemoryVariationShortPerformancePct
+	case term == "medium_term" && engine == "cost":
+		return s.CPUVariationMediumCostPct, s.MemoryVariationMediumCostPct
+	case term == "medium_term" && engine == "performance":
+		return s.CPUVariationMediumPerformancePct, s.MemoryVariationMediumPerformancePct
```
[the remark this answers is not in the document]
upadhyeammit: This is good to refactor, however I will raise another separate pull request for this.

migrations/000024_recommendation_sets_add_current_request_columns.up.sql
```diff
@@ -0,0 +1,6 @@
+-- Add current request columns to recommendation_sets for sorting (mirrors namespace_recommendation_sets).
+-- Existing rows remain NULL; the poller fills values for new records.
+-- List queries use ORDER BY ... DESC NULLS LAST (see listoptions.SQLOrderByFragment).
+ALTER TABLE recommendation_sets
+    ADD COLUMN cpu_request_current FLOAT,
+    ADD COLUMN memory_request_current FLOAT;
```
saltgen: Let's switch to Numeric type with previously discussed bounds.

internal/model/recommendation_set.go
```diff
@@ -15,10 +15,28 @@ import (
...
+	CPUVariationMediumCostPct           float64 `gorm:"column:cpu_variation_medium_cost_pct;type:numeric(10,4)"`
+	CPUVariationMediumPerformancePct    float64 `gorm:"column:cpu_variation_medium_performance_pct;type:numeric(10,4)"`
+	CPUVariationLongCostPct             float64 `gorm:"column:cpu_variation_long_cost_pct;type:numeric(10,4)"`
+	CPUVariationLongPerformancePct      float64 `gorm:"column:cpu_variation_long_performance_pct;type:numeric(10,4)"`
+	MemoryVariationShortCostPct         float64 `gorm:"column:memory_variation_short_cost_pct;type:numeric(10,4)"`
+	MemoryVariationShortPerformancePct  float64 `gorm:"column:memory_variation_short_performance_pct;type:numeric(10,4)"`
+	MemoryVariationMediumCostPct        float64 `gorm:"column:memory_variation_medium_cost_pct;type:numeric(10,4)"`
+	MemoryVariationMediumPerformancePct float64 `gorm:"column:memory_variation_medium_performance_pct;type:numeric(10,4)"`
+	MemoryVariationLongCostPct          float64 `gorm:"column:memory_variation_long_cost_pct;type:numeric(10,4)"`
+	MemoryVariationLongPerformancePct   float64 `gorm:"column:memory_variation_long_performance_pct;type:numeric(10,4)"`
```
saltgen: Worth considering: these variation fields might be better as *float64 instead of float64.

When a term doesn't have enough monitoring data yet (e.g. medium_term needs ~7 days but only 2 exist), the field defaults to Go's 0.0 and gets written to the DB. This puts "no data" rows right in the middle of the sort:

```
DESC NULLS LAST (current — float64):
  50.0   ← increase recommended
  25.0
   0.0   ← no data, masquerading as "no change"
  -5.0   ← decrease recommended
 -10.0
  NULL   ← old db default init rows

DESC NULLS LAST (with *float64):
  50.0
  25.0
  -5.0
 -10.0
  NULL   ← old rows + no-data rows, all at the bottom
```

With *float64, unset fields stay nil, GORM writes NULL, and they sort last where they belong. The read side (StoredVariationPcts) already uses *float64 and handles nil correctly, so no changes needed there.

Same applies to NamespaceRecommendationSet — that's pre-existing but need to be cleaned up together.

internal/model/common.go
```diff
@@ -15,10 +15,64 @@ const (
...
+		s.CPUVariationMediumPerformancePct != nil ||
+		s.CPUVariationLongCostPct != nil ||
+		s.CPUVariationLongPerformancePct != nil ||
+		s.MemoryVariationShortCostPct != nil ||
+		s.MemoryVariationShortPerformancePct != nil ||
+		s.MemoryVariationMediumCostPct != nil ||
+		s.MemoryVariationMediumPerformancePct != nil ||
+		s.MemoryVariationLongCostPct != nil ||
+		s.MemoryVariationLongPerformancePct != nil
+}
```
saltgen, line 54:
```suggestion
func (s *StoredVariationPcts) HasValues() bool {
	return s.CPUVariationShortCostPct != nil
}
```

Since the poller writes all 12 fields in a single transaction (all succeed or all roll back), a single field check is sufficient because partial column population is not possible.
Feel free to add this as comment for context.

internal/api/handlers.go
```diff
@@ -126,7 +127,9 @@ func GetRecommendationSet(c echo.Context) error {
 			recommendationSet.ClusterUUID,
 			unitChoices,
 			setk8sUnits,
-			recommendationSet.Recommendations)
+			recommendationSet.Recommendations,
+			nil, // single-item fetch: recompute variation percentages from JSON
```
saltgen: Let's use pre-computed StoredVariationPcts here instead of recomputing, unless I am missing something here.

internal/api/handlers.go
```diff
@@ -240,7 +244,9 @@ func GetNamespaceRecommendationSet(c echo.Context) error {
 			nsRecommendationSet.ClusterUUID,
 			unitChoices,
 			setk8sUnits,
-			nsRecommendationSet.Recommendations)
+			nsRecommendationSet.Recommendations,
+			nil, // single-item fetch: recompute variation percentages from JSON
```
saltgen: https://github.com/RedHatInsights/ros-ocp-backend/pull/626/changes#r3130117653

saltgen: https://github.com/RedHatInsights/ros-ocp-backend/pull/626/changes#r3130117653

internal/api/utils.go
```diff
@@ -869,7 +926,13 @@ func UpdateRecommendationJSON(handlerName string, recommendationID string, clust
...
 	data = transformComponentUnits(unitsToTransform, updateUnitsk8s, data) // cpu: core values require truncation
 	data = filterNotifications(recommendationID, clusterUUID, data)
-	data = convertVariationToPercentage(data)
+
+	if storedPcts != nil && storedPcts.HasValues() {
+		data = injectStoredRequestVariationPct(data, storedPcts)
+		data = convertVariationToPercentage(data, true) // limits only; requests already injected
+	} else {
+		data = convertVariationToPercentage(data, false)
+	}
```
saltgen:
```suggestion
skipRequests := storedPcts != nil && storedPcts.HasValues()
if skipRequests {
    data = injectStoredRequestVariationPct(data, storedPcts)
}
data = convertVariationToPercentage(data, skipRequests)
```

saltgen (changes requested):
Hello Amit,

A couple of items,

- Response schemas in openapi json for variation has format as units, request you to look into the same and other relevant aspects for consistency.


- Nit, out of scope for this PR: SQLOrderByFragment appends NULLS LAST on every DESC query, including non-nullable columns like container_name. It should only apply to nullable columns (cpu_request_current, memory_request_current, and the *_variation_*_pct columns) where pre-migration rows have NULLs. Harmless on PostgreSQL today but worth tightening up separately.

internal/model/common.go
upadhyeammit: This can create false negative: e.g. cpu_variation_short_cost_pct is NULL but memory_variation_medium_cost_pct (or any other) is set. HasValues() becomes false, you never inject, you always recompute requests from JSON. That’s usually still correct for the client, but you stop using the DB for those other columns even though they’re there.

saltgen (approved):
Hello @upadhyeammit,

Let's add a couple of unittests for `injectStoredRequestVariationPct` function in the refactor PR.

It seems to be core logic for these changes, that said I don't see a need to block this PR, thanks!

saltgen (approved):
unittests ++ 
Thanks @upadhyeammit

Aurobinda55:
@upadhyeammit, I see some api tests failures related to containers. I checked those failures and found . No error in the ROS processor pod; it validates the request, and the CSV is sent for experiments. But there is no data in the recommendation table. Also, I see an error in the recommendation-poller ("error", "max_endtime_report":"2026-04-27T09:00:00Z","msg":"unable to save a record into recommendation set). Nothing changed recently that will break the form from IQE. Can you please check the backend side?

upadhyeammit:
> @upadhyeammit, I see some api tests failures related to containers. I checked those failures and found . No error in the ROS processor pod; it validates the request, and the CSV is sent for experiments. But there is no data in the recommendation table. Also, I see an error in the recommendation-poller ("error", "max_endtime_report":"2026-04-27T09:00:00Z","msg":"unable to save a record into recommendation set). Nothing changed recently that will break the form from IQE. Can you please check the backend side?

I have added a protecting code block to avoid this situation and also aligned the columns to float as per namespace table for container table, this should work now

Aurobinda55:
> > @upadhyeammit, I see some api tests failures related to containers. I checked those failures and found . No error in the ROS processor pod; it validates the request, and the CSV is sent for experiments. But there is no data in the recommendation table. Also, I see an error in the recommendation-poller ("error", "max_endtime_report":"2026-04-27T09:00:00Z","msg":"unable to save a record into recommendation set). Nothing changed recently that will break the form from IQE. Can you please check the backend side?
> 
> I have added a protecting code block to avoid this situation and also aligned the columns to float as per namespace table for container table, this should work now

- Smoke tests looks green 
- Newly added sorting tests looks good for different variation including current and limit 
- No regression on Namespace functionality , tests are looks good 

PR can be merged now.Thanks!

migrations/000024_recommendation_sets_add_current_request_columns.up.sql
```diff
@@ -2,5 +2,5 @@
 -- Existing rows remain NULL; the poller fills values for new records.
 -- List queries use ORDER BY ... DESC NULLS LAST (see listoptions.SQLOrderByFragment).
 ALTER TABLE recommendation_sets
-    ADD COLUMN cpu_request_current NUMERIC(10, 4),
-    ADD COLUMN memory_request_current NUMERIC(10, 4);
+    ADD COLUMN cpu_request_current FLOAT,
+    ADD COLUMN memory_request_current FLOAT;
```
saltgen: @upadhyeammit Do we expect issues for this type wrt variation columns?

upadhyeammit: yeah... in edge cases (e.g., CPU current request of 0.001 cores with a variation of 100 cores), the percentage would be 10,000,000%  which would overflow NUMERIC(10,4). The clamp caps it to 999,999.9999

internal/utils/conversion.go
```diff
@@ -21,6 +24,33 @@ func clampToNumeric10_4Range(p float64) float64 {
...
+
+// ClampToNumeric10_4 clamps a value into the PostgreSQL NUMERIC(10,4) range.
+func ClampToNumeric10_4(v float64) float64 {
+	return clampToNumeric10_4Range(v)
+}
+
+// ClampToNumeric20_4 clamps a value into the PostgreSQL NUMERIC(20,4) range.
+func ClampToNumeric20_4(v float64) float64 {
+	return clampToNumeric20_4Range(v)
+}
```
saltgen: Is it possible to uppercase the actual functions instead of using the exportable wrappers?

saltgen (changes requested):

saltgen (approved):

PR #657, opened 2026-05-19
RHINENG-26485 Fix Unleash config overwrite

saltgen:

Unleash Clowder config was getting overwritten by local, since it was present outside the else block.

- [ ] Yes
- [x] No

## Security Checklist :lock:

- [x] Bugfix
- [ ] New Feature
- [ ] Refactor
- [ ] Unittests Added
- [ ] DRY code
- [ ] Dependency Added
- [ ] DB Migration Added

## Additional :mega:

Changes tested locally i.e. non-EE with Kruize 0.10 and performance profile v2, able to see namespace and container recommendations.

### DIFF
--- a/internal/config/config.go
+++ b/internal/config/config.go
@@ -211,7 +211,7 @@ func initConfig() {
 
 		// Unleash config
 		viper.SetDefault("UnleashClientAccessToken", "rosocp:dev.token")
-		viper.SetDefault("UnleashHostname", "0.0.0.0")
+		viper.SetDefault("UnleashHostname", "localhost")
 		viper.SetDefault("UnleashScheme", "http")
 		viper.SetDefault("UnleashPort", 3063)
 		viper.SetDefault(
--- a/internal/config/config.go
+++ b/internal/config/config.go
@@ -208,6 +208,20 @@ func initConfig() {
 
 		// Sources-api-go
 		viper.SetDefault("SOURCES_API_BASE_URL", "http://127.0.0.1:8002")
+
+		// Unleash config
+		viper.SetDefault("UnleashClientAccessToken", "rosocp:dev.token")
+		viper.SetDefault("UnleashHostname", "0.0.0.0")
+		viper.SetDefault("UnleashScheme", "http")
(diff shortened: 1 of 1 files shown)


internal/config/config.go
```diff
@@ -201,6 +201,20 @@ func initConfig() {
 
 		// Sources-api-go
 		viper.SetDefault("SOURCES_API_BASE_URL", "http://127.0.0.1:8002")
+
+		// Unleash config
```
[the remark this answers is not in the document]
saltgen: Theoretical concern

internal/config/config.go
```diff
@@ -201,6 +201,20 @@ func initConfig() {
...
+		viper.SetDefault("UnleashClientAccessToken", "rosocp:dev.token")
+		viper.SetDefault("UnleashHostname", "0.0.0.0")
```
upadhyeammit:
```suggestion
		viper.SetDefault("UnleashHostname", "localhost")
```

upadhyeammit (changes requested):

saltgen:
@Aurobinda55 This is ready to test on manual EE

upadhyeammit (approved):

internal/config/config.go
r14chandra: The PR doesn’t actually resolve the original issue. The Unleash SetDefault calls outside the if/else block are still present, and those outside-block defaults always overwrite whatever was set inside either the if or else block. As a result, in the Clowder path, the values are still being overwritten by the hardcoded defaults.

As you test locally, it doesn’t catch this because both the else block and the outside block resolve to the same local Unleash server.

The fix should be to move the Unleash defaults into the else block and remove the duplicate defaults from outside the if/else entirely.

r14chandra (changes requested):

r14chandra (approved):

Aurobinda55:
Changes looks good from QE side . 
on an EE deployment , added the unleash configuration through UI  and the org_id in filter list . Verified that the recommendation are generated for that org_id and skip for the other id's.

upadhyeammit (approved):

